PR #65, opened 2022-04-28
Mono repo init

TheOrangePuff:
This pulls all our CDK repos into one mono repo. The main purpose of this it to setup the structure to make our repos easier to maintain in the future. Especially in regards to upgrading everything from CDK 1 to CDK 2.

At the moment this does nothing too fancy, it's purely a collection of all our CDK repos. It allows an `npm ci` and `npm build` to run without errors (although there are warnings) and adds dependabot checks for each package.

There are two packages that already use CDK 2. As CDK 1 and 2 and not compatible I've excluded the CDK 2 repos in the tsconfig. Once we upgrade everything else to CDK 2 we can remove these from the exclude list. 

Once this is merged I will need to go through each CDK repo and pull in any changes that have been made since I created this repo. Then I will archive the old repos.

I had a very ~~painful~~ fun time fixing git history on this so I hope the branch history is okay :crossed_fingers: 

`DO-1284: Code Review`


crispy101:
Check the packages that I wrote (rds alert, rabbitmq, waf, pipeline) and looks good. Daniel has done a great work here. 

cdk-pipeline seems to have quite a big change in CDK v2.0, so we may want to refactor the project it was used for.

crispy101 (approved):
LGTM

PR #111, opened 2022-05-05
Add release for basic auth

TheOrangePuff:
Can't test the GitHub release locally so am merging this in to test. I will add the other ones assuming this release works.

As we spoke about in guild there will be one release pipe for each package. I've tried to make it as copy paste friendly as possible by using a package variable.


TheOrangePuff:
Wait this won't work at all... Ignore me

PR #151, opened 2022-05-18
include missing lambda:ListTags

krishanthisera:
`lambda:ListTags` This will allow lambda to list function tags. 
for docs: [see](https://docs.aws.amazon.com/lambda/latest/dg/API_ListTags.html)


krishanthisera:
Closing this  as @TheOrangePuff has already addressed this

PR #243, opened 2022-06-06
Upgrade to cdk v2

TheOrangePuff:
Lots of changes in this one!

Unsurprisingly CDK 2 depreciates / changes some functions so there are a few changes to the code. Luckily most deprecated functions have a simple drop in replacement. (e.g. `fromLookup` becomes `fromLookupById`).

CDK 2 also changes the way imports are handled. Therefore every import statement that was referencing CDK has been updated. 

The major breaking change (from what I've noticed) is the way tests are handled. Fortunately, we haven't written many tests so the upgrade process for this repo was pretty easy. (CDK Stacks contain a few more tests so will be more challenging). 

A few other minor changes:
- Moved all packages into root `package.json` (upgrading to CDK 2 meant all cdk packages are common between constructs)
- Changed dependabot to reflect the above change
- Added an `npm run test` step to the build pipeline


packages/basic-auth/lib/basic-auth-construct.ts
```diff
@@ -19,6 +19,8 @@ export class BasicAuthFunction extends Construct {
             this,
             'BasicAuthFunction',
             {
+              runtime: Runtime.NODEJS_12_X,
```
AdamJHall, line 22: We should probably update all of these runtimes to node 16 like our serverless projects. Did we want to do that as part of this upgrade or we can create a separate ticket?

AdamJHall (approved):
Looks good! Just one comment about the runtimes :)

packages/basic-auth/lib/basic-auth-construct.ts
TheOrangePuff: Yeah I noticed this too! I was going to add it to our audit sheet, thanks for the reminder :smile:. I'll make the changes in this repo as part of another ticket. (I figured upgrading everything to CDK 2 was already a huge change so I didn't want to throw in a node runtime change too)

PR #249, opened 2022-06-14
Include lambda:PutFunctionConcurrency action to the service role policy

krishanthisera:
This PR allows  `lambda:PutFunctionConcurrency` action. This is required, if you are intended to use `Reserved concurrency` when setting up the your Lambdas.


krishanthisera:
Closing as this will address in another PR

PR #290, opened 2022-07-01
DO-1285: Added the Export names for cfnOutput

krishanthisera:
Introduced a parameter call `stackExportPrefix` and enable export name in CloudFormation output.

### DIFF
--- a/packages/static-hosting/lib/static-hosting.ts
+++ b/packages/static-hosting/lib/static-hosting.ts
@@ -42,7 +42,7 @@ export class StaticHosting extends Construct {
     constructor(scope: Construct, id: string, props: StaticHostingProps) {
         super(scope, id);
         
-        // Should the stackExportPrefix is empty it should StaticHosting as the prefix
+        // Should the stackExportPrefix is empty, 'StaticHosting' should be used as the prefix 
         const exportPrefix =  props.stackExportPrefix ? props.stackExportPrefix :  'StaticHosting'
 
         const siteName = `${props.subDomainName}.${props.domainName}`;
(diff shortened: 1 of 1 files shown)


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -63,6 +67,7 @@ export class StaticHosting extends Construct {
             new CfnOutput(this, 'S3LoggingBucketName', {
                 description: "S3 Logs",
                 value: s3LoggingBucket.bucketName,
+                exportName: `${exportPrefix}S3LoggingBucketName`
```
TheOrangePuff, line 70: What was the default value of exportName before this change? As in if I upgrade my static hosting stack to this version, will all my resources get re-created with a different name? Ideally we don't want this behaviour.

I'm assuming that's what `const exportPrefix =  props.stackExportPrefix ? props.stackExportPrefix :  'StaticHosting'` is for? If so, maybe add a comment saying _why_ the default value is set to that just to make it clear.

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -7,6 +7,7 @@ import { CloudFrontTarget } from 'aws-cdk-lib/aws-route53-targets';
 import { User, Group, Policy, PolicyStatement, Effect } from 'aws-cdk-lib/aws-iam';
 
 export interface StaticHostingProps {
+    stackExportPrefix?: string,
```
TheOrangePuff, line 10: This is a little nit-picky but I'd probably call this `exportPrefix` as this is a construct not a stack.

TheOrangePuff (changes requested):
Just a couple things to look at :slightly_smiling_face:

packages/static-hosting/lib/static-hosting.ts
krishanthisera: This is a good call I would change that.

packages/static-hosting/lib/static-hosting.ts
krishanthisera: There were no default value, it means it is empty. 
The exportPrefix is used, in case if you got two of these stack in a same region, there will be a conflict, if you should search the same export name.

TheOrangePuff: Right, that makes sense :slightly_smiling_face:

TheOrangePuff (approved):
LGTM

PR #311, opened 2022-07-04
DO-1285: S3 Extended props

krishanthisera:
There are some cases, where we might need to pass additional props rather than defaults for S3 bucket.
For instance,  we might need to pass a Life Cycle Rule for bucket Objects. 
This PR intended to accommodate such a scenario.


TheOrangePuff (approved):
LGTM

PR #312, opened 2022-07-08
Feature/GitHub Release workflow

krishanthisera:
This PR is intend to reinvent the Release workflow, for CDK-Constructs.

Sample release tags, which could trigger this workflow:
- static-hosting-1.0.0
- static-hosting-1.0.0-beta


AdamJHall:
@krishanthisera,

Looks pretty good, could you also add release process instructions to the readme as well? Just so it's documented easily somewhere :)

TheOrangePuff (changes requested):
Looks great! Like @AdamJHall said, just add something to the README and it'll be ready for merge.

AdamJHall (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
+name: Publish to NPM
+
+on:
+  release:
+    tags: 
+      - '**-[0-9].[0-9].[0-9]'
+      - '**-[0-9].[0-9].[0-9]-*'
```
TheOrangePuff, line 7: @krishanthisera I just noticed this. Not sure if actions follow standard regex? But this if it does, this will only capture single digits e.g. `1.2.1` but not `1.12.10`

krishanthisera: Ohh yeah good spot. I think we may have to fix this :)

PR #313, opened 2022-07-11
Feature/do 1285 monorepo workflow configs

krishanthisera:
In this PR

- It has been found that when publishing packages, `private: true`, make them skip.
- Update the `tsconfig.json` to follow monorepo approach.


TheOrangePuff (approved):
Looks good! I thought I already changed the `tsconfig.json` to reference the root package but I might be misremembering...

PR #330, opened 2022-07-12
DO-1342: Prerender proxy internal

TheOrangePuff:
Part one of hosting Prerender internally. 

This is almost a direct copy of the original `prerender-proxy` construct with a few major differences:
 - Supports a custom URL
 - Works with basic auth (instead of a Prerender Token)

Note that this is CDK 1 as CDK 2 doesn't support bundling (yet).


packages/prerender-proxy/lib/handlers/prerender.ts
```diff
@@ -1,42 +1,49 @@
...
-                 domainName: 'service.prerender.io',
-                 port: 443,
-                 protocol: 'https',
-                 readTimeout: 20,
-                 keepaliveTimeout: 5,
-                 sslProtocols: ['TLSv1', 'TLSv1.1', 'TLSv1.2'],
-                 path: '/https%3A%2F%2F' + request.headers['x-prerender-host'][0].value,
-                 customHeaders: {
-                      'x-prerender-token': [{
-                           key: 'x-prerender-token',
```
krishanthisera, line 31: @TheOrangePuff  Shouldn't we maintain prerenderToken? As if this to be used in existing implementations

krishanthisera (commented):
LGTM. :clinking_glasses: Just one concern as commented

packages/prerender-proxy/lib/handlers/prerender.ts
TheOrangePuff: That was a good pickup. I wasn't meant to commit those files... I'll revert that

krishanthisera (approved):

PR #331, opened 2022-07-12
DO-1342: Prerender hosted in fargate

TheOrangePuff:
Contains code to deploy Prerender as a Fargate service. 

With this configuration Prerender is behind basic auth and uses an s3 bucket as storage for rendered pages.

As this construct was built using CDK 2, it is being merged into `feature/DO-1285-upgrade-cdk`.


packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        // Create a load-balanced Fargate service 
+        const fargateService = new ecsPatterns.ApplicationLoadBalancedFargateService(
+            this,
+            `${props.prerenderName}-service`,
+            {
+                cluster,
+                serviceName: `${props.prerenderName}-service`,
+                desiredCount: 1,
+                cpu: 512, // 0.5 vCPU (this may need to be increased)
```
krishanthisera, line 65: How about if we pass, resource limits as props. Rather than hard-coding it?

krishanthisera (changes requested):
Good work @TheOrangePuff :clinking_glasses: Just one concern as I commented

packages/prerender-fargate/lib/prerender-fargate.ts
TheOrangePuff: Good idea. I might add these as defaults but allow a manual override :+1:

krishanthisera (approved):

PR #368, opened 2022-07-20
DO-1360: Change prerender fargate to use a token

TheOrangePuff:
This is part of the process to combine the prerender proxy functions into one. 

This replaces the previous basic auth method with a token method. This brings it inline with what prerender.io uses.


AdamJHall (approved):
Looks good!

PR #369, opened 2022-07-20
DO-1360: Combine prerender proxy constructs

TheOrangePuff:
This removes the internal prerender proxy and instead provides the option of providing a custom url for prerender. 

With the change to the prerender fargate construct here: https://github.com/aligent/cdk-constructs/pull/368, we can use the token authentication method for both internal and prerender.io services.


AdamJHall (approved):
Looks good :) Good idea on the tsconfigs

PR #382, opened 2022-07-22
Prerender CloudFront cache control

krishanthisera:
CloudFront Cache control function for Prerender.
This function to be associated with Origin Response


packages/prerender-proxy/README.md
```diff
@@ -7,7 +8,13 @@ The `prerender` will function also make a HEAD request to a nominated backend to
 
 These functions are intended to be added to an existing Cloudfront  
 
+## Cache Control
+
+The intention of the `cache-control` function is to avoid/control CloudFront Caches for `prenrender` bot.
```
TheOrangePuff, line 13: Small typo: `prenrender`

packages/prerender-proxy/lib/handlers/cache-control.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    if (response.headers[`${cacheKey}`]) {
+        response.headers['Cache-Control'] = [
+            {
+                key: 'Cache-Control',
+                value: `max-age=${cacheMaxAge}`
+            }
+        ]
+    }
+    console.log(JSON.stringify(response.headers));
```
TheOrangePuff, line 20: This should be removed, not ideal to log headers in production :slightly_smiling_face:

packages/prerender-proxy/README.md
```diff
@@ -7,7 +8,13 @@ The `prerender` will function also make a HEAD request to a nominated backend to
...
+This function to be associated with CloudFront's`origin response`
+
 ## Props
+
 `redirectBackendOrigin`: The backend origin to make the HEAD request to.
 `redirectFrontendHost`: This hostname is used to replace the backend host for any redirects that contain the backend host.
 `prerenderToken`: Your prerender.io authentication token
```
TheOrangePuff, line 20: Can you add the new env variables to the README? `PRERENDER_CACHE_KEY`, `PRERENDER_CACHE_MAX_AGE` and add a note that they are optional.

packages/prerender-proxy/index.ts
```diff
@@ -2,5 +2,6 @@ import { PrerenderLambda } from "./lib/prerender-lambda-construct";
 import { PrerenderFunction } from "./lib/prerender-construct";
 import { PrerenderCheckFunction } from "./lib/prerender-check-construct";
 import { ErrorResponseFunction } from "./lib/error-response-construct";
+import { CloudFrontCacheControl } from "./lib/prerender-cf-cache-control-consctruct"
```
TheOrangePuff, line 5: Type in file name: `./lib/prerender-cf-cache-control-consctruct` should be `./lib/prerender-cf-cache-control-construct`

packages/prerender-proxy/lib/prerender-lambda-construct.ts
```diff
@@ -23,5 +26,6 @@ export class PrerenderLambda extends Construct {
     
     this.errorResponseFunction = new ErrorResponseFunction(this, 'ErrorResponse', {});
 
+    this.cacheControlFunction = new CloudFrontCacheControl(this, 'PrerenderCloudFrontCacheControl', props.cacheControlProps)
```
TheOrangePuff, line 29: Should we have an option to disable this function?

TheOrangePuff (changes requested):
Couple of changes for you :slightly_smiling_face:

packages/prerender-proxy/lib/prerender-lambda-construct.ts
krishanthisera: Good question, should we?  As I understand this is a flaw?

packages/prerender-proxy/lib/handlers/cache-control.ts
krishanthisera: Yeah that should be gone

packages/prerender-proxy/lib/prerender-lambda-construct.ts
TheOrangePuff: I'm just thinking for existing builds. Always nice to have the option to fall back to what we previously had. Although going forward I feel like most builds will implement this. 
I just like options :)

krishanthisera: Hmm, for the existing implementations, it will only create the function but you will have to explicitly introduce this when you use this construct. 
Should this be a problem?

TheOrangePuff (approved):

packages/prerender-proxy/lib/prerender-lambda-construct.ts
TheOrangePuff: For future reference, we had a discussion about this and as only creates the lambda function and doesn't automatically add it to the behaviour it's fine to not have an option to disable.

PR #392, opened 2022-07-25
ERO-1280: Robots and sitemap remap

TheOrangePuff:
Adds the ability to remap paths from both S3 and the backend host.

Some common use cases would be:
 - Mapping `robots.txt` in the PWA to `robots.txt` in the backend. 
 - Mapping `robots.txt` in the PWA to `robots-production.txt` in the S3 bucket.

This PR also allows you to optionally disable enforceSSL.


gowrizrh:
image

:+1: :smile:

packages/static-hosting/lib/handlers/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
gowrizrh, line 3: remove this if no scripst are setup?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,9 +1,11 @@
...
 import { Bucket, BucketEncryption, BlockPublicAccess } from '@aws-cdk/aws-s3';
-import { OriginAccessIdentity, CloudFrontWebDistribution, PriceClass, ViewerProtocolPolicy, SecurityPolicyProtocol, SSLMethod, Behavior, SourceConfiguration, CloudFrontWebDistributionProps } from '@aws-cdk/aws-cloudfront';
+import { OriginAccessIdentity, CloudFrontWebDistribution, PriceClass, ViewerProtocolPolicy, SecurityPolicyProtocol, SSLMethod, Behavior, SourceConfiguration, CloudFrontWebDistributionProps, LambdaFunctionAssociation, LambdaEdgeEventType } from '@aws-cdk/aws-cloudfront';
 import { HostedZone, RecordTarget, ARecord } from '@aws-cdk/aws-route53';
 import { CloudFrontTarget } from '@aws-cdk/aws-route53-targets';
 import { User, Group, Policy, PolicyStatement, Effect } from '@aws-cdk/aws-iam';
+import { Version } from '@aws-cdk/aws-lambda';
+import { ArbitraryPathRemapFunction } from './arbitrary-path-remap';
 
 export interface StaticHostingProps {
```
gowrizrh, line 10: you could look to just extend the props from the parent construct so you don't need to redefine a lot of the properties that already are defined

packages/static-hosting/lib/handlers/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "lib": ["es2019"],
+    "removeComments": true,
+    "moduleResolution": "node",
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "sourceMap": true,
+    "target": "es2017",
```
gowrizrh, line 9: should the target and lib match? i've no clue

gowrizrh (commented):
i've just added a few basic questions

packages/static-hosting/lib/handlers/package.json
TheOrangePuff: I just copied that from the other handlers :sweat_smile: I can remove it though

packages/static-hosting/lib/handlers/tsconfig.json
TheOrangePuff: They don't _need_ to match. My understanding is that lib is what you're using when writing TS whereas target is what version of JS it transpiles to.

packages/static-hosting/lib/static-hosting.ts
TheOrangePuff: Not sure that this would help much? I don't think any of the properties defined in `StaticHostingProps` are properties of `Construct`

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -221,4 +258,23 @@ export class StaticHosting extends Construct {
             });
         };
     };
+
+    private CreateRemapBehavior(from: string, to: string): Behavior {
```
gowrizrh, line 262: camelCase for function names?

gowrizrh (approved):
LGTM. minor code style suggestion

PR #394, opened 2022-07-26
DO-1362: Map static files to S3 origin by default

TheOrangePuff:
By default static files should be mapped to the S3 bucket. 

In a situation where this is not desired, there is an optional parameter to disable the behaviour.


crispy101 (commented):
LGTM

crispy101 (approved):

gowrizrh:
LGTM

PR #401, opened 2022-07-27
Update the main file in package.json

krishanthisera:
Update main in the package.json. this should fix, `SyntaxError: Cannot use import statement outside a module` issue.


TheOrangePuff (approved):
LGTM

PR #438, opened 2022-08-15
Include cdk exports

krishanthisera:
This PR is CDK V1 version of #290 

> Introduced a parameter call stackExportPrefix and enable export name in CloudFormation output.

### DIFF
--- a/packages/static-hosting/lib/static-hosting.ts
+++ b/packages/static-hosting/lib/static-hosting.ts
@@ -8,7 +8,7 @@ import { Version } from '@aws-cdk/aws-lambda';
 import { ArbitraryPathRemapFunction } from './arbitrary-path-remap';
 
 export interface StaticHostingProps {
-    stackExportPrefix?: string, 
+    exportPrefix?: string, 
     domainName: string;
     subDomainName: string;
     certificateArn: string;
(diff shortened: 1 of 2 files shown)


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -53,6 +54,9 @@ export class StaticHosting extends Construct {
 
     constructor(scope: Construct, id: string, props: StaticHostingProps) {
         super(scope, id);
+        
+        // Should the stackExportPrefix is empty, 'StaticHosting' should be used as the prefix 
+        const exportPrefix =  props.exportPrefix ? props.exportPrefix :  'StaticHosting'
```
TheOrangePuff, line 59: What's the advantage of having this default to `StaticHosting` rather than just a blank string?

TheOrangePuff (changes requested):
Just had a question

packages/static-hosting/lib/static-hosting.ts
krishanthisera: Well this is act as an identifier, say if it is blank string: The bucket name would be exported as `BucketName`, which perhaps you may have already using `BucketName` as an export in one of your other stack.

TheOrangePuff: Oh okay that makes sense :slightly_smiling_face:

TheOrangePuff (approved):

PR #502, opened 2022-08-29
DO-1363: Resolve default vpc

TheOrangePuff:
Allow a vpc id to be optionally passed. In the event one is not provided, use the default vpc.


gowrizrh (approved):
LGTM

PR #686, opened 2022-10-31
Updated WAF Construct

AdamJHall:
A client is looking to add the WAF to their Cloudfront distribution so I have made some small refactors to make this a little easier in the upstream repo. The main changes are just making a few more props optional and allowing a list of associations to be provided. 

This is a breaking change so there will need to be a major version bump.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5839,28 +5839,10 @@
                 "constructs": "^3.3.69"
             }
         },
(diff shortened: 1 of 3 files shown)


TheOrangePuff (approved):
LGTM :)

crispy101 (approved):
LGTM. Thanks for tidying this up.

PR #712, opened 2022-11-08
Update CDK to 1.180.0

gowrizrh:
Also adds a tiny hack to enable http3

https://github.com/aligent/cdk-constructs/pull/712/files#diff-75f7e6cacfbbc4e1202de58b3d7d614bcd5b6cd7e34cc1392d0f9b81d7548f98R260


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -255,6 +257,7 @@ export class StaticHosting extends Construct {
         if (props.enableErrorConfig) {
             distributionProps = {
                 ...distributionProps, ...{
+                    httpVersion: 'http3' as HttpVersion.HTTP2,
```
TheOrangePuff, line 260: I love this work around haha But can you add a TODO here to fix this once we upgrade to cdk 2?

TheOrangePuff (approved):
Looks good! Just one small comment

PR #714, opened 2022-11-10
Cloudfront WAF Support

AdamJHall:
This PR adds support for the WAF to be used with our Cloudfront distribution so we can lockdown via IP, add AWS WAF rules etc.

Changes:
Added support to the WAF Construct to allow both REGIONAL and CLOUDFRONT scopes
Added support to the static hosting construct to import a Web ACL


TheOrangePuff (approved):
LGTM

gowrizrh (approved):

PR #715, opened 2022-11-10
Update WAF Construct to CDK 1.180.0

AdamJHall:
Update some WAF dependencies to 1.180.0. 

@gowrizrh, Were these intentionally skipped? Just want to make sure I'm not going to break it :D


gowrizrh (approved):
Not at all! I must have missed it.

AdamJHall:
Looks like the build failed due to a difference between the pipeline node version and the nvmrc version. I've fixed that up in the latest commit and updated the package-lock.json to match.

PR #879, opened 2023-01-11
Add Shared VPC construct

toddhainsworth:
Takes https://bitbucket.org/aligent/aligent-shared-vpc-stack/ and turns it into a CDK construct.

### DIFF
--- a/packages/shared-vpc/lib/shared-vpc.ts
+++ b/packages/shared-vpc/lib/shared-vpc.ts
@@ -24,10 +24,9 @@ export interface SharedVpcProps {
    */
   hostedZoneDomain?: string,
   /**
-   * The optional zone records as a JSON string
-   * Ex: "[{ "type": "A", "target": "10.6.0.12", "record": "subdomain" }]"
+   * Optional zone records
    */
-  hostedZoneRecords?: string,
+  hostedZoneRecords?: Zone[],
 }
 
 export class SharedVpc extends Construct {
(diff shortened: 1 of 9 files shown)


toddhainsworth:
Rrr it's added as a submodule :facepalm: 
I'll see if I can revert that and redo the commit.

packages/shared-vpc/lib/shared-vpc.ts
```diff
@@ -0,0 +1,125 @@
...
+  cidr?: string,
+  /**
+   * The optional domain to use for the hosted-zone
+   */
+  hostedZoneDomain?: string,
+  /**
+   * The optional zone records as a JSON string
+   * Ex: "[{ "type": "A", "target": "10.6.0.12", "record": "subdomain" }]"
+   */
+  hostedZoneRecords?: string,
```
AdamJHall, line 30: Can we just make this a `Zone[]` ? Instead of a json string and parsing later?

toddhainsworth: Yeah nice, since we're not accepting this from the CLI anymore, we don't need to ask for a JSON string :pray:

TheOrangePuff (dismissed):
Looks good to me :tada:

AdamJHall (approved):

toddhainsworth:
One more commit, sorry :sweat_smile: I forgot to update the README with the Zone type instead of JSON string

toddhainsworth:
One more commit, sorry :sweat_smile: I forgot to update the README with the Zone type instead of JSON string

TheOrangePuff (approved):
LGTM

PR #939, opened 2023-02-02
DO-1438: Use middified handlers with the prerender proxy

krishanthisera:
### Use middified handlers with the prerender proxy

- We have recently pulled out edge handlers from the prerender proxy as [lambda-at-edge-handlers](https://github.com/aligent/cdk-constructs/tree/main/packages/lambda-at-edge-handlers) 
- Here, we are associating those edge functions with the help of middy
- Ideally, we shall pull these [handler functions](https://github.com/aligent/cdk-constructs/tree/main/packages/lambda-at-edge-handlers) from this repo to a separate mono-repo dedicated to edge handlers

### DIFF
--- a/packages/prerender-proxy/lib/handlers/cache-control.ts
+++ b/packages/prerender-proxy/lib/handlers/cache-control.ts
@@ -1,21 +1,8 @@
-import 'source-map-support/register';
-import { CloudFrontResponseEvent, CloudFrontResponse  } from 'aws-lambda';
-/*  
-    Prerender cache control function 
-    Consider associate this function as *origin response* function
-*/
-export const handler = async (event: CloudFrontResponseEvent): Promise<CloudFrontResponse> => {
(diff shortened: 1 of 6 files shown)


packages/prerender-proxy/lib/handlers/cache-control.ts
```diff
@@ -1,21 +1,8 @@
...
-            }
-        ]
-    }
-    return response;
-}
+/**
+ * Prerender Cache Control Handler
+ * @link https://github.com/aligent/cdk-constructs/blob/main/packages/lambda-at-edge-handlers/lib/cache-control.ts
+ */
+export const handler = middy(PrerenderCacheControlHandler);
```
gowrizrh, line 8: Instead of returning the functions as a middy wrapper, perhaps we must make this into a middy middleware instead? That means end user can make their own base functions and wrap it with middy, and then call `.use(prerenderMiddleWare)`

TheOrangePuff:
Stale branch, closing

PR #959, opened 2023-02-06
Migrate static hosting to CDK v2

gowrizrh:
I've removed all constructs (to be ported individually) so that there's no painful merge conflicts.

Please review files under `packages/static-hosting`.


TheOrangePuff:
@gowrizrh while we are in the process of upgrading to CDK 2, we may still need to update the old constructs. Can this instead target an epic branch based off of main? Then once everything is go to go with CDK 2 we can merge that epic branch into main

PR #1042, opened 2023-03-28
Revert "JBH-615: Set is a bot header in prerender proxy repo"

krishanthisera:
Reverting the PR as this could be achieved using `x-prerender`, `x-request-prerender`


krishanthisera:
Proposed change was carried out as a part of  #1051

PR #1046, opened 2023-05-22
fix: process authentication first

TheOrangePuff:
By processing authentication first, we fix the issue where the s3cache would kick off first on `/`. This would result in showing the contents of the S3 bucket rather than a 401.


gowrizrh (approved):
LGMT

PR #1047, opened 2023-05-22
fix: set healthcheck to success on 401 status

TheOrangePuff:


packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -94,6 +94,13 @@ export class PrerenderFargate extends Construct {
             }
         );
 
+        fargateService.targetGroup.configureHealthCheck({
+            path: "/health",
+            interval: Duration.seconds(120),
+            unhealthyThresholdCount: 5,
+            healthyHttpCodes: '401'
```
gowrizrh, line 101: A comment explaining this would be helpful :pray:

PR #1051, opened 2023-05-26
Prerender check function code cleanup

krishanthisera:
## Prerender check function code cleanup

We have introduced following header to address some concerns with GeoIP redirect logic. 
As we are no longer required this header we shall remove it. 

**Further note:**
This header was used to check if a request coming from a crawler, and if it does, it should NOT redirect based on the GeoIP logic. Instead of using this header, we found that we could leverage,    
-  `x-prerender`
- `x-request-prerender`
headers to see it is a crawler.

### DIFF
--- a/packages/prerender-proxy/lib/handlers/prerender-check.ts
+++ b/packages/prerender-proxy/lib/handlers/prerender-check.ts
@@ -15,12 +15,6 @@ export const handler = async (
   // then set the x-request-prerender header so the origin-request lambda function
   // alters the origin to prerender.io
   if (IS_BOT.test(request.headers["user-agent"][0].value)) {
-    request.headers["x-is-a-bot"] = [
-      {
-        key: "x-is-a-bot",
-        value: "true",
-      },
-    ];
     if (!IS_FILE.test(request.uri) && !request.headers["x-prerender"]) {
       request.headers["x-request-prerender"] = [
         {


packages/prerender-proxy/lib/handlers/prerender-check.ts
```diff
@@ -15,12 +15,6 @@ export const handler = async (
   // then set the x-request-prerender header so the origin-request lambda function
   // alters the origin to prerender.io
   if (IS_BOT.test(request.headers["user-agent"][0].value)) {
```
gowrizrh, line 17: On a side note I'm wondering whether this function is needed at all.

CloudFront behaviour can just simply forward all headers which means this function is not necessary at the viewer request right.

Simply move the existing logic in `prerender.ts` under this condition and we save invoking the viewer request lambda?

gowrizrh (approved):

packages/prerender-proxy/lib/handlers/prerender-check.ts
gowrizrh: @AdamJHall @TheOrangePuff @crispy101 Thoughts?

AdamJHall: Is the user agent part of the cache key though? Otherwise we would end up serving the prerender response to everyone.

gowrizrh: True but that would be controlled outside of this construct wouldn't it?

gowrizrh: And that's not really supported in CDK v1 anyway

> If you want to include headers in the cache key, use a cache policy. For more information, see [Creating cache policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/controlling-the-cache-key.html#cache-key-create-cache-policy) in the Amazon CloudFront Developer Guide .

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront.CfnDistribution.ForwardedValuesProperty.html#headers

gowrizrh: https://github.com/aws/aws-cdk/issues/10829

PR #1062, opened 2023-07-31
DO-1531: Migrate static hosting to CDK v2

gowrizrh:

* Delete packages and migrate `static-hosting` 

Only relevant changeset are these two files for now
image

Changes to other packages can be iteratively migrated upon. I didn't want to open one giant PR with changes in every package.


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,144 +1,99 @@
...
+  Bucket,
+  BucketEncryption,
+  BucketProps,
+} from "aws-cdk-lib/aws-s3";
 import { CSP } from "../types/csp";
-import { RecordTarget } from "@aws-cdk/aws-route53";
-import { CloudFrontTarget } from "@aws-cdk/aws-route53-targets";
 
 export interface StaticHostingProps {
+  domainNames: string[];
```
TheOrangePuff, line 42: Are we concerned about backwards compatibility? I think it's fine not to worry but we need a migration guide

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,144 +1,99 @@
...
-  behaviors?: Array<Behavior>;
-  enableErrorConfig?: boolean;
-  enableStaticFileRemap?: boolean;
-  remapPaths?: remapPath[];
-  backendHost?: string;
-  remapBackendPaths?: remapPath[];
+  errorResponsePagePath?: string;
+  webAclId?: string;
   defaultRootObject?: string;
-  enforceSSL?: boolean;
```
TheOrangePuff, line 65: Why remove these options?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,144 +1,99 @@
...
-   * The main purpose of this csp header is to provide a method of setting a report-uri.
-   */
+  certificateArn: string;
+  defaultBehaviorEdgeLambdas: EdgeLambda[];
+  additionalBehaviors?: Record<string, BehaviorOptions>;
+  enableErrorConfig?: boolean;
+  createDnsRecord?: boolean;
+  zoneName?: string;
   csp?: CSP;
-  /**
```
TheOrangePuff, line 75: Not sure why the comments have been removed?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,144 +1,99 @@
...
 export interface ResponseHeaderMappings {
-  header: ResponseHeadersPolicy;
-  pathPatterns: string[];
-  attachToDefault?: boolean;
+  defaultBehaviorResponseHeaderPolicy?: ResponseHeadersPolicy;
+  additionalBehaviorResponsePolicy?: Record<string, ResponseHeadersPolicy>;
 }
 
-export class StaticHosting extends Construct {
-  private staticFiles = [
```
TheOrangePuff, line 109: Has this functionality been removed entirely?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -221,226 +170,107 @@ export class StaticHosting extends Construct {
...
+    );
 
-    let originConfigs = new Array<SourceConfiguration>();
+    const originCachePolicy = new CachePolicy(this, "S3OriginCachePolicy", {
+      headerBehavior: CacheHeaderBehavior.allowList("x-forwarded-host"),
+      enableAcceptEncodingBrotli: true,
+      enableAcceptEncodingGzip: true,
+    });
 
-    // Add the backend host as an origin
```
TheOrangePuff, line 237: This is a feature I use all the time, why are we removing it?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -1,144 +1,99 @@
...
     const exportPrefix = props.exportPrefix
       ? props.exportPrefix
       : "StaticHosting";
 
-    const siteName = `${props.subDomainName}.${props.domainName}`;
-    const siteNameArray: Array<string> = [siteName];
+    const siteName = props.domainNames[0];
     const enforceSSL = props.enforceSSL !== false;
-    const enableStaticFileRemap = props.enableStaticFileRemap !== false;
     const disableCSP = props.disableCSP === true;
```
TheOrangePuff, line 95: This was removed from props so it'll always be false right?

TheOrangePuff (changes requested):
This is a huge change so docs need to be updated. Can you update the README please? :pray: 

I've probably missed a bit because this is a lot to process :sweat_smile: Updating docs will help heaps!

packages/static-hosting/lib/static-hosting.ts
gowrizrh: I think it's still present in the props. Rearranging might have made it look like it's been removed

packages/static-hosting/lib/static-hosting.ts
gowrizrh: Yeah, because behaviors are to be added outside of static hosting but I'm investigating alternative options to keep this in. It will have to be part of a future PR

packages/static-hosting/lib/static-hosting.ts
gowrizrh: The option still exists, see other comment. I'll mark this as resolved.

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -221,226 +170,107 @@ export class StaticHosting extends Construct {
...
 
-    // Create behaviors to map static content to bucket
-    if (enableStaticFileRemap) {
-      for (const path of this.staticFiles) {
-        originConfigs[originConfigs.length - 1].behaviors.push(
-          this.createRemapBehavior(`*.${path}`, `*.${path}`)
-        );
-      }
-    }
+    const additionalBehaviors: Record<string, Writeable<BehaviorOptions>> = {};
```
AdamJHall, line 234: How do you add additionalBehaviors? It doesn't look like we are using the props here

AdamJHall (commented):
I know this is just an upgrade but while we are here can we add class properties of the underlying resources (e.g. s3 buckets, distribiton, etc). That way if we need to modify them outside the construct we can.

packages/static-hosting/lib/static-hosting.ts
TheOrangePuff: That would be great. This is vital to our hosting infra so definitely needed!

gowrizrh: After playing around, we may not require this. The only reason we have this is we create them as additional behaviours so that these behaviours can be excluded from having any lambda association. This was a workaround / hack because of the one lambda limitation.

This is a solved problem now since we just let the function decide whether assets need to pass through or not.

packages/static-hosting/lib/static-hosting.ts
gowrizrh: I've retained these :sweat_smile:

packages/static-hosting/lib/static-hosting.ts
gowrizrh: we're defining `additionalBehaviors` and then taking the `reponseHeaderPolicies` (prop name retanied from v1 for easier migration) and just passing it as is.

if additional behaviours are defined per path, it's just iterated over and added as `additionalBehaviors` to the distributions API.

packages/static-hosting/lib/static-hosting.ts
gowrizrh: This has now been added back.

packages/static-hosting/lib/static-hosting.ts
gowrizrh: misread your comment, this is now done via merging the `additionalBehaviors` and `props.additionalBehaviors`.

TheOrangePuff (approved):
Looks good to me, but probably best to get another DO to review too :pray:

PR #1068, opened 2023-09-13
Add read only user to access ecs

TheOrangePuff:

* Adds a role to allow users from a different account to assume a role that grants access to view ECS and relevant log information. 

**Notes to reviewers**  

### DIFF
--- a/packages/graphql-mesh-server/lib/fargate.ts
+++ b/packages/graphql-mesh-server/lib/fargate.ts
@@ -1,5 +1,5 @@
 import { Construct } from 'constructs';
-import { Duration, Token } from 'aws-cdk-lib';
+import { CfnOutput, Duration } from 'aws-cdk-lib';
 import { RemovalPolicy } from 'aws-cdk-lib';
 import * as acm from 'aws-cdk-lib/aws-certificatemanager';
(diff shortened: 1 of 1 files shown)


packages/graphql-mesh-server/lib/fargate.ts
```diff
@@ -216,5 +233,57 @@ export class MeshService extends Construct {
             ],
             adjustmentType: auto_scaling.AdjustmentType.CHANGE_IN_CAPACITY,
         });
+
+        // Cross account role to get read only access to mesh
+        // and the relevant logs
+        if (props.readOnlyRoleName) {
```
AdamJHall, line 239: This should probably be `(props.readOnlyRoleName && props.awsAccountArn)` OR just check `props.awsAccountArn` and set a default for `props.readOnlyRoleName `

AdamJHall (changes requested):

packages/graphql-mesh-server/lib/fargate.ts
TheOrangePuff: @AdamJHall done! Ready for another round

AdamJHall (approved):

PR #1069, opened 2023-09-15
Feature/do 1484 enable cache for redirects

crispy101:

* Let Prerender be able to cache 301 and 302 responses in the S3 bucket when explicitly enabled by enableRedirectCache flag 

* cached object meta-tags for this purpose
image

* N/A

**Notes to PR author**

**Notes to reviewers**  
Incorporated the below existing plug-ins
* prerender.httpHeaders: https://github.com/prerender/prerender/blob/master/lib/plugins/httpHeaders.js
* S3Cache: https://github.com/prerender/prerender-aws-s3-cache/blob/master/index.js

### DIFF
--- a/packages/prerender-fargate/lib/prerender/server.js
+++ b/packages/prerender-fargate/lib/prerender/server.js
@@ -41,7 +41,7 @@ if (process.env.ENABLE_REDIRECT_CACHE.toLowerCase() === 'true'){
     var he = require('he');
     var s3 = new (require('aws-sdk')).S3({params:{Bucket: process.env.S3_BUCKET_NAME}});
     server.use({
-            requestReceived: function(req, res, next) {
+        requestReceived: function(req, res, next) {
                 if(req.method !== 'GET' && req.method !== 'HEAD') {
                     console.log("skipping requestReceived from S3 Cache... ")
                     return next();
@@ -54,11 +54,10 @@ if (process.env.ENABLE_REDIRECT_CACHE.toLowerCase() === 'true'){
                 }
 
                 s3.getObject({
-                        Key: key
+                    Key: key
                 }, function (err, result) {
 
                     if (!err && result) {
-                        console.log(result.Metadata);
                         console.log("Found cached object: " + key);
                         if (result.Metadata.location){
                             res.setHeader('Location', result.Metadata.location);
@@ -72,7 +71,6 @@ if (process.env.ENABLE_REDIRECT_CACHE.toLowerCase() === 'true'){
                 });
             },
             // The pageLoaded function is a modified version of https://github.com/prerender/prerender/blob/478fa6d0a5196ea29c88c69e64e72eb5507b6d2c/lib/plugins/httpHeaders.js
-
             pageLoaded: function(req, res, next) {
                 const statusCodesToCache = ['200', '301', '302'];
                 let metaTagStatusCode = 200;
--- a/packages/prerender-fargate/lib/prerender/server.js
+++ b/packages/prerender-fargate/lib/prerender/server.js
@@ -36,8 +36,114 @@ server.use({
 });
 
 server.use(prerender.blacklist());
-server.use(prerender.httpHeaders());
-server.use(prerender.removeScriptTags());
-server.use(s3Cache);
+
+if (process.env.ENABLE_CACHE_FOR_REDIRECTION){
+    server.use(prerender.removeScriptTags());
+    var he = require('he');
+    var s3 = new (require('aws-sdk')).S3({params:{Bucket: process.env.S3_BUCKET_NAME}});
+    server.use({
+            requestReceived: function(req, res, next) {
+                if(req.method !== 'GET' && req.method !== 'HEAD') {
+                    console.log("skipping requestReceived from S3 Cache... ")
+                    return next();
+                }
+    
+                var key = req.prerender.url;
+    
+                if (process.env.S3_PREFIX_KEY) {
+                    key = process.env.S3_PREFIX_KEY + '/' + key;
+                }
+    
+                s3.getObject({
+                        Key: key
+                }, function (err, result) {
+    
+                    if (!err && result) {
+                        console.log(result.Metadata);
+                        console.log("Found cached object: " + key);
+                        if (result.Metadata.location){
+                            res.setHeader('Location', result.Metadata.location);
+                        }
+                        return res.send(result.Metadata.httpreturncode, result.Body);
+                    } else {
+                        console.error(err);
+                    }
+    
+                    next();
+                });
+            },
+    
+            pageLoaded: function(req, res, next) {
+                const statusCodesToCache = ['200', '301', '302'];
+                let metaTagStatusCode = 200;
+                let location = '';
+    
+                if (req.prerender.content && req.prerender.renderType == 'html') {
+                    let statusMatch = /<meta[^<>]*(?:name=['"]prerender-status-code['"][^<>]*content=['"]([0-9]{3})['"]|content=['"]([0-9]{3})['"][^<>]*name=['"]prerender-status-code['"])[^<>]*>/i,
+                    headerMatch = /<meta[^<>]*(?:name=['"]prerender-header['"][^<>]*content=['"]([^'"]*?): ?([^'"]*?)['"]|content=['"]([^'"]*?): ?([^'"]*?)['"][^<>]*name=['"]prerender-header['"])[^<>]*>/gi,
+                    head = req.prerender.content.toString().split('</head>', 1).pop(),
+                    // statusCode = 200,
+                    match;
+    
+                    if (match = statusMatch.exec(head)) {
+                        metaTagStatusCode = match[1] || match[2];
+                        req.prerender.content = req.prerender.content.toString().replace(match[0], '');
+                        console.log("metaTagStatusCode: " + metaTagStatusCode);
+                    }
+    
+                    while (match = headerMatch.exec(head)) {
+                        location = he.decode(match[2] || match[4]);
+                        res.setHeader(match[1] || match[3], location);
+                        req.prerender.content = req.prerender.content.toString().replace(match[0], '');
+                    }
+    
+                    // Skip caching for the http response codes not in the list, such as 404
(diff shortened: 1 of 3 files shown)


packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                }, function(err, result) {
+                    console.log(result);
+                    if (err) console.error(err);
+
+                    next();
+                });
+        }
+    });
+} else {
+    server.use(prerender.httpHeaders());
```
TheOrangePuff, line 144: Why remove:
```
server.use(prerender.httpHeaders());
```
When caching 301s and 302s?

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
 });
 
 server.use(prerender.blacklist());
-server.use(prerender.httpHeaders());
-server.use(prerender.removeScriptTags());
-server.use(s3Cache);
+
+if (process.env.ENABLE_REDIRECT_CACHE.toLowerCase() === 'true'){
+    server.use(prerender.removeScriptTags());
```
TheOrangePuff, line 41: This is used in both conditions so it can be pulled out

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+    var he = require('he');
+    var s3 = new (require('aws-sdk')).S3({params:{Bucket: process.env.S3_BUCKET_NAME}});
+    server.use({
+            requestReceived: function(req, res, next) {
+                if(req.method !== 'GET' && req.method !== 'HEAD') {
+                    console.log("skipping requestReceived from S3 Cache... ")
```
TheOrangePuff, line 47: Minor point but I would remove this console log. It could be useful during local dev but in a prod system it wouldn't be very helpful (e.g. no logging of which request is being skipped).

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                if (process.env.S3_PREFIX_KEY) {
+                    key = process.env.S3_PREFIX_KEY + '/' + key;
+                }
+
+                s3.getObject({
+                        Key: key
+                }, function (err, result) {
+
+                    if (!err && result) {
+                        console.log(result.Metadata);
```
TheOrangePuff, line 62: This console log is also questionable in terms of debugging usefulness, I'd probably remove it too

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                const statusCodesToCache = ['200', '301', '302'];
+                let metaTagStatusCode = 200;
+                let location = '';
+
+                if (req.prerender.content && req.prerender.renderType == 'html') {
+                    let statusMatch = /<meta[^<>]*(?:name=['"]prerender-status-code['"][^<>]*content=['"]([0-9]{3})['"]|content=['"]([0-9]{3})['"][^<>]*name=['"]prerender-status-code['"])[^<>]*>/i,
+                    headerMatch = /<meta[^<>]*(?:name=['"]prerender-header['"][^<>]*content=['"]([^'"]*?): ?([^'"]*?)['"]|content=['"]([^'"]*?): ?([^'"]*?)['"][^<>]*name=['"]prerender-header['"])[^<>]*>/gi,
+                    head = req.prerender.content.toString().split('</head>', 1).pop(),
+                    // statusCode = 200,
+                    match;
```
TheOrangePuff, line 86: Just for clarity I'd put a `let` on each line rather than using the comma syntax. I can be a bit hard to read especially with the long regexes :sweat_smile:

TheOrangePuff: Also if a variable isn't going to change you should use the `const` keyword rather than `let`

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+
+                    if (match = statusMatch.exec(head)) {
```
TheOrangePuff, line 88: I'd split this into two lines. Having a variable assignment in an if statement makes it easy to misinterpret what the logic is

TheOrangePuff: Also it might be clearer with something like this:
```
const statusMatchRegex = /regexhere/;
const statusMatch = statusMatchRegex.exec(head);

if (statusMatch) {
 ....
```
Then similar case for headerMatch as well

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                        metaTagStatusCode = match[1] || match[2];
```
TheOrangePuff, line 89: Would be worth adding some comments for what the various `match` indexes mean. You can kinda infer it from the code but this feels like the perfect place for a comment :slightly_smiling_face:

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                        req.prerender.content = req.prerender.content.toString().replace(match[0], '');
+                        console.log("metaTagStatusCode: " + metaTagStatusCode);
```
TheOrangePuff, line 91: Again this console log isn't super helpful

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                    }
+
+                    while (match = headerMatch.exec(head)) {
```
TheOrangePuff, line 94: Why is this in a `while` loop while the one above is an `if` statement? :thinking:

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                        location = he.decode(match[2] || match[4]);
+                        res.setHeader(match[1] || match[3], location);
+                        req.prerender.content = req.prerender.content.toString().replace(match[0], '');
+                    }
+
+                    // Skip caching for the http response codes not in the list, such as 404
+                    if ( ! statusCodesToCache.includes(metaTagStatusCode.toString()) ) {
+                        console.log("metaTagStatusCode " + metaTagStatusCode + " is not in the cachable code list. Returning without caching.");
```
TheOrangePuff, line 102: Probably not a useful log either. Also fun JS syntax fact, it's generally preferred to use string literals rather than `+`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

This would become: 
```
console.log(`metaTagStatusCode ${metaTagStatusCode} is not in the cachable code list. Returning without caching.`);
```

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                    }
+                }
+
+                if(req.prerender.statusCode !== 200) {
+                    return next();
+                }
+
+                // Override req.prerender.statusCode with the StatusCode returned via the meta tag.
+                // If metaTagStatusCode is not in the statusCodesToCache array or req.prerender.statusCode is not 200, then this line wouldn't be reached. Therefore no if condition for this overriding is needed.
+                req.prerender.statusCode = metaTagStatusCode;
```
TheOrangePuff, line 113: What's the purpose of the `metaTagStatusCode` variable, why not just assign to `req.prerender.statusCode` directly?

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+                console.log("Caching the object with statusCode " + req.prerender.statusCode);
```
TheOrangePuff, line 114: Might be worth stating which object is being cached (url key maybe?)

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,114 @@ server.use({
...
+
+                var key = req.prerender.url;
+                var s3Metadata = {
+                    httpreturncode: req.prerender.statusCode.toString()
+                }
+
+                if (location) {
```
TheOrangePuff, line 121: I feel like this could be done earlier on, it feels a bit out of place referencing the location variable here. I reckon you could change this so `location` is scoped inside an existing `if` statement above rather than being scoped to the function

TheOrangePuff (changes requested):
Quite a few comments here :sweat_smile: Happy to chat anything through of you have any questions. Logic is good just mostly stylistic comments :slightly_smiling_face:

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -36,8 +36,102 @@ server.use({
...
+if (process.env.ENABLE_REDIRECT_CACHE.toLowerCase() === 'true'){
+    var he = require('he');
+    var s3 = new (require('aws-sdk')).S3({params:{Bucket: process.env.S3_BUCKET_NAME}});
+    server.use({
+            // The requestReceived and pageLoaded functions are a modified version of
+            // httpHeader plugin - https://github.com/prerender/prerender/blob/478fa6d0a5196ea29c88c69e64e72eb5507b6d2c/lib/plugins/httpHeaders.js combined with
+            // s3cache plugin - https://github.com/prerender/prerender-aws-s3-cache/blob/98707fa0f787de83aa41583682cd2c2d330a9cca/index.js
+        requestReceived: function(req, res, next) {
+                if(req.method !== 'GET' && req.method !== 'HEAD') {
+\                    return next();
```
TheOrangePuff, line 49: I don't think this slash should be here :thinking:

packages/prerender-fargate/lib/prerender/server.js
TheOrangePuff: Resolve, this was implemented as part of the cache step.

TheOrangePuff (changes requested):
Looking good! One minor comment

TheOrangePuff (approved):
LGTM :tada:

PR #1072, opened 2023-09-26
Prerender fargate vpc setup

krishanthisera:
------

* Create the VPC by default instead of using the default VPC

* N/A

* N/A

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/prerender-fargate/lib/prerender-fargate.ts
+++ b/packages/prerender-fargate/lib/prerender-fargate.ts
@@ -10,20 +10,63 @@ import { AccessKey, User } from 'aws-cdk-lib/aws-iam';
 import { Duration, RemovalPolicy, Stack } from 'aws-cdk-lib';
 import * as path from 'path';
 
+/**
+ * Options for configuring a Prerender Fargate service.
+ */
 export interface PrerenderOptions {
-    prerenderName: string,
-    domainName: string,
-    vpcId?: string,
-    bucketName?: string,
-    expirationDays?: number,
-    tokenList: Array<string>,
-    certificateArn: string,
-    desiredInstanceCount?: number,
-    maxInstanceCount?: number,
-    instanceCPU?: number,
-    instanceMemory?: number
-    enableRedirectCache?: string
-}
+    /**
+     * The name of the Prerender service.
+     */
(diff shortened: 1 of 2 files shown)


packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -10,20 +10,63 @@ import { AccessKey, User } from 'aws-cdk-lib/aws-iam';
...
+     */
+    instanceMemory?: number;
+    /** 
+     * Whether to enable caching of redirects in the Prerender service. 
+     */
+    enableRedirectCache?: string;
+    /** 
+     * Whether to create a new VPC for the Prerender service. 
+     */
+    createVpc: boolean;
```
TheOrangePuff, line 68: I don't think we need this, `vpcId` is optional so if it's not provided we should create a VPC

TheOrangePuff (changes requested):

packages/prerender-fargate/lib/prerender-fargate.ts
krishanthisera: This is intentional, this would ensure we won't accidentally create a VPC again for existing stacks.

packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -52,8 +95,26 @@ export class PrerenderFargate extends Construct {
...
+                        subnetType: ec2.SubnetType.PUBLIC,
+                    },
+                    {
+                        name: 'private',
+                        subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS,
+                    },
+                ],
+            });
+        } else {
+            const vpcLookup = { vpcId: props.vpcId };
```
TheOrangePuff, line 115: Can we add the isDefault logic here?

packages/prerender-fargate/lib/prerender-fargate.ts
TheOrangePuff: That makes sense!

TheOrangePuff (changes requested):

crispy101:
Invalid. Closing off.

PR #1074, opened 2023-09-29
Minor cleanup

gowrizrh:

* Remove unused test
* Remove `any` type definition
* Change `prepublish` command from `tsc` to `npm run build`
* Add `@aligent/esbuild` as a dependency to packages that use it.


TheOrangePuff (dismissed):
LGTM

packages/prerender-proxy/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Cloudfront Lambda@Edge constructs for integrating with prerender.io",
   "main": "index.js",
   "scripts": {
-    "build": "tsc && cd ./lib/handlers && npm ci",
-    "prepublish": "tsc && cd ./lib/handlers && npm ci"
```
TheOrangePuff, line 8: Not sure about this one... Do we not need to install in the handler directory?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ES2018",
     "module": "commonjs",
-    "lib": ["es2018"],
+    "lib": ["es2018", "DOM"],
```
TheOrangePuff, line 5: What does this do?

TheOrangePuff (commented):
Couple questions

tsconfig.json
gowrizrh: This is to mitigate https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/826

Esbuild requires the WASM module

packages/prerender-proxy/package.json
gowrizrh: This is because we are no longer building it "outside" of CDK. When we're running Esbuild from within CDK, the packages are available from the construct's `package.json`

This let's CDK handle lambda functions as first class citizens instead of treating it as an asset that is built outside and bundled in cdk.

TheOrangePuff (approved):

PR #1106, opened 2023-10-11
Epic/do 1545 prerender fargate improvment

krishanthisera:
------

* Enable S3 endpoint as an optional feature
* Code cleanup and documentation
* Update TSConfig globally:
```json
...
   "declaration": false,
    "noEmit": true,
...
```

* 

* 

**Notes to PR author**

**Notes to reviewers**  


tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "target": "ES2018",
     "module": "commonjs",
     "lib": ["es2018"],
-    "declaration": true,
+    "declaration": false,
+    "noEmit": true,
```
TheOrangePuff, line 7: Why are we enabling noEmit?

packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -114,5 +115,16 @@ export class PrerenderFargate extends Construct {
             scaleInCooldown: Duration.seconds(60),
             scaleOutCooldown: Duration.seconds(60),
         });
+
+        /**
+         * Enable VPC Endpoints for S3
+         * This would  create S3 endpoints in all the PUBLIC subnets of the VPC
+         */
+        if (props.enableS3Endpoint) {
```
TheOrangePuff: Should this default to true?

TheOrangePuff (commented):
Couple of questions :)

packages/prerender-fargate/lib/prerender-fargate.ts
krishanthisera: This would be associated with a cost so it should be optional.

tsconfig.json
krishanthisera: The idea is we don't need compiler output for the CDK constructs.

packages/prerender-fargate/lib/prerender-fargate.ts
TheOrangePuff: But doesn't it make it cheaper? Surely that should be a default

crispy101: As long as it's backward-compatible, which it is AFAIK, I'm happy with making this default, too.

crispy101 (approved):
LGTM

packages/prerender-fargate/lib/prerender-fargate.ts
krishanthisera: I'm happy to make this as default, but this would NOT make anything cheaper. You would have a running cost (per GB) and cost of having the endpoint (~0.01 USD X number of subnets).

tsconfig.json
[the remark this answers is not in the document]
krishanthisera: When you run `npm ci` this creates compiler output which is not convenient. CDK should handle the compiled outputs.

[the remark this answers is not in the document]
AdamJHall: If we don't emit type declarations then when you install the module you won't get the types for your props / interfaces will you?

PR #1109, opened 2023-10-13
Mesh Updates

AdamJHall:
------

This PR updates the mesh construct to use the VPC interface so that we can import existing vpcs  if required. I have also added the database config value for redis so that the mesh can use a shared redis instance with different databases for each. 

**Notes to PR author**

**Notes to reviewers**  


AdamJHall:
Looks like I missed the slack notifications with the last PR, it was sitting on another branch. I've merged that here as well to bring it all inline.

packages/graphql-mesh-server/lib/pipeline.ts
```diff
@@ -90,5 +104,44 @@ export class CodePipelineService extends Construct {
         }),
       ],
     });
+
+    if (props.notificationArn) {
+      const notifier = new NodejsFunction(this, "NotifierLambda", {
+        entry: path.resolve(__dirname, "../assets/handlers/notify-sns.ts"),
```
gowrizrh, line 110: @AdamJHall this would upload the ts file without compilation to js right?

AdamJHall: CDK auto detects that it is Typescript and transpiles it itself based on your global tsconfig

gowrizrh: :open_mouth: Does it also bundle and transpile? If the function has a dependency or a package will it be included in the function? (excluding `aws-sdk` because these are available in the runtime already)

AdamJHall: I believe so but not 100%, it just uses esbuild under the hood so it should.

AdamJHall: Ahh yes it does: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs-readme.html#install-modules

gowrizrh (dismissed):

TheOrangePuff (dismissed):

AdamJHall:
@gowrizrh and @TheOrangePuff can you please re-review? There were some merge conflicts that needed to be fixed up.

gowrizrh (approved):

PR #1211, opened 2023-11-16
feat: Graphql Mesh add WAF rules

TheOrangePuff:

* Add default rate limiting rule
* Allow passing of custom WAF rules

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/graphql-mesh-server/lib/fargate.ts
+++ b/packages/graphql-mesh-server/lib/fargate.ts
@@ -64,9 +64,9 @@ export interface MeshServiceProps {
   blockedIps?: string[];
   /**
    * The waf rule priority.
-   * Defaults to 2 
+   * Defaults to 2
    */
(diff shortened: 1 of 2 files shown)


krishanthisera (approved):
LGTM. Minor suggestions for further improvement:
- We could separate waf configuration from props to a separate struct and give more control over waf configuration. 
- Let the implementation stack decide whether to enable WAF but the default is WAF enable. There could be instances that we would want to use Firewall Manager.

TheOrangePuff:
> LGTM. Minor suggestions for further improvement:
> 
>     * We could separate waf configuration from props to a separate struct and give more control over waf configuration.
> 
>     * Let the implementation stack decide whether to enable WAF but the default is WAF enable. There could be instances that we would want to use Firewall Manager.

Yep makes sense! Happy to see this as a future improvement :slightly_smiling_face:

PR #1216, opened 2023-11-17
feat: enable container insights

TheOrangePuff:

* Enable container insights by default

**Notes to reviewers**  

### DIFF
--- a/packages/graphql-mesh-server/lib/fargate.ts
+++ b/packages/graphql-mesh-server/lib/fargate.ts
@@ -181,7 +181,8 @@ export class MeshService extends Construct {
 
     const cluster = new ecs.Cluster(this, `cluster`, {
       vpc: this.vpc,
-      containerInsights: props.containerInsights || true,
+      containerInsights:
+        props.containerInsights !== undefined ? props.containerInsights : true,
     });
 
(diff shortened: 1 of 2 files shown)


packages/graphql-mesh-server/lib/fargate.ts
```diff
@@ -176,6 +181,7 @@ export class MeshService extends Construct {
 
     const cluster = new ecs.Cluster(this, `cluster`, {
       vpc: this.vpc,
+      containerInsights: props.containerInsights || true,
```
AdamJHall, line 184: This will always evaluate to true, you have to do:
```
containerinsights: props.containerInsights !== undefined ?  props.containerInsights :  true,
```

Otherwise you can never turn it off

AdamJHall (changes requested):

packages/graphql-mesh-server/lib/fargate.ts
TheOrangePuff: Oh, of course :disappointed:

TheOrangePuff: Fixed :slightly_smiling_face:

AdamJHall (approved):

PR #1268, opened 2023-12-11
HOTFIX: Fix Prerender proxy lambda@edge

AdamJHall:
------

This PR allows passing through a request and cache policy so that prerender headers can be added to the allowlist. Otherwise the prerender headers are not passed to the origin request lambda@edge function and the origin switch to prerender is not performed.

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/static-hosting/lib/static-hosting.ts
+++ b/packages/static-hosting/lib/static-hosting.ts
@@ -108,9 +108,13 @@ export interface StaticHostingProps {
   overrideLogicalId?: string;
 
   /**
-   *
+   * A Request policy used on the default behavior
    */
   defaultBehaviorRequestPolicy?: OriginRequestPolicy;
+
+  /**
+   * A Cache policy used on the default behavior
+   */
   defaultBehaviorCachePolicy?: CachePolicy;
 }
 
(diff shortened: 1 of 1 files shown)


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -255,20 +265,20 @@ export class StaticHosting extends Construct {
...
-    );
+      });
 
-    const originCachePolicy = new CachePolicy(this, "S3OriginCachePolicy", {
-      headerBehavior: CacheHeaderBehavior.allowList("x-forwarded-host"),
-      enableAcceptEncodingBrotli: true,
-      enableAcceptEncodingGzip: true,
-    });
+    const originCachePolicy =
+      props.defaultBehaviorCachePolicy ||
```
gowrizrh, line 276: Should we parametrise the `allowList` or the entire object? Either way is fine, passing an array for allowed headers is what I had in mind.

gowrizrh (approved):

packages/static-hosting/lib/static-hosting.ts
AdamJHall: I had that initially but figured this might be better so that we can override other properties on the policies later (cookies, query strings etc).

PR #1299, opened 2024-02-02
Fix/DO-1610: address prematured completion of recache requests

crispy101:

* Use `for` instead of `forEach` in order to ensure each of the URLs are requested before the lambda funtion completes its execution
* Add more meaningful logs

* 

* 

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/prerender-fargate/lib/recaching/prerender-recache-api-construct.consumer.ts
+++ b/packages/prerender-fargate/lib/recaching/prerender-recache-api-construct.consumer.ts
@@ -1,4 +1,4 @@
-import { Context, SQSEvent, SQSRecord } from "aws-lambda";
+import { Context, SQSEvent } from "aws-lambda";
 import axios from "axios";
 
 const userAgent = "prerender / Googlebot recaching request";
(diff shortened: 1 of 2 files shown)


packages/prerender-fargate/lib/recaching/prerender-recache-api-construct.consumer.ts
```diff
@@ -9,13 +9,18 @@ const userAgent = "prerender / Googlebot recaching request";
...
-    await axios.get(url, {
+    const res = await axios.get(url, {
       headers: {
         "User-Agent": userAgent,
       },
     });
-  });
+    console.log(
+      `Response data: ${JSON.stringify(
+        res.data
```
TheOrangePuff, line 22: Logging the entire response data is going to produce some pretty big logs. Same with headers. I don't think we need that all the time. Can we just log the status code and the url fetched?

TheOrangePuff (changes requested):

packages/prerender-fargate/lib/recaching/prerender-recache-api-construct.consumer.ts
crispy101: Thanks. Yep, updated. We, however, may want to have a debugging flag in an update later.

TheOrangePuff (approved):

PR #1312, opened 2024-02-18
feat: add lambda to pipe to invalidate cache

TheOrangePuff:

* Adds an optional lambda function that invalidates the `/graphql` endpoint on a successful mesh deployment 

**Notes to reviewers**  

### DIFF
--- a/packages/graphql-mesh-server/assets/handlers/invalidate-cloudfront-cache.ts
+++ b/packages/graphql-mesh-server/assets/handlers/invalidate-cloudfront-cache.ts
@@ -1,0 +1,29 @@
+import {
+  CloudFrontClient,
+  CreateInvalidationCommand,
+  CreateInvalidationCommandInput,
+  CreateInvalidationCommandOutput,
+} from "@aws-sdk/client-cloudfront";
+
+const cfClient = new CloudFrontClient();
+
+const PATHS = process.env.PATHS as string;
+const DISTRIBUTION_ID = process.env.DISTRIBUTION_ID as string;
+
+export const handler = async (): Promise<CreateInvalidationCommandOutput> => {
+  const paths = PATHS.split(",");
+
+  const invalidationInput: CreateInvalidationCommandInput = {
+    DistributionId: DISTRIBUTION_ID,
+    InvalidationBatch: {
+      Paths: {
+        Quantity: paths.length,
+        Items: paths,
+      },
+      CallerReference: "lambda",
+    },
+  };
+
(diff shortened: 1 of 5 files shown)


packages/graphql-mesh-server/assets/handlers/invalidate-cloudfront-cache.ts
```diff
@@ -0,0 +1,29 @@
...
+      Paths: {
+        Quantity: paths.length,
+        Items: paths,
+      },
+      CallerReference: "lambda",
+    },
+  };
+
+  const command = new CreateInvalidationCommand(invalidationInput);
+  return await cfClient.send(command);
```
AdamJHall, line 28: If you're returning the result you don't need to await the result

packages/graphql-mesh-server/lib/pipeline.ts
```diff
@@ -110,6 +115,49 @@ export class CodePipelineService extends Construct {
...
+            __dirname,
+            "../assets/handlers/invalidate-cloudfront-cache.ts"
+          ),
+          description: "Lambda function to invalidate CloudFront cache.",
+          runtime: Runtime.NODEJS_18_X,
+          handler: "index.handler",
+          timeout: Duration.seconds(5),
+          environment: {
+            PATHS: "/graphql",
+            DISTRIBUTION_ID: props.cloudFrontDistributionId || "",
```
AdamJHall, line 133: You don't need the `||` here as you are checking for `cloudFrontDistributionId` above

AdamJHall (dismissed):
Couple of small changes that could be made but don't really affect the result so happy to approve this now :)

packages/graphql-mesh-server/assets/handlers/invalidate-cloudfront-cache.ts
TheOrangePuff: Oh really? I was worried if I didn't await the lambda might not finish sending the command

packages/graphql-mesh-server/lib/pipeline.ts
TheOrangePuff: Ah yeah, must have added this before I included the if statement

packages/graphql-mesh-server/assets/handlers/invalidate-cloudfront-cache.ts
AdamJHall: Nah the function itself is async, so there is an implied await on the result (Could be wrong here but that's my understanding).

AdamJHall (approved):

PR #1353, opened 2024-05-03
Add optional props for construct name

gowrizrh:

* Adds optional names to ECS cluster, service and repository, looks more readable in the AWS console
* Removes the any IP4v security group rule in favour for service's security group (the other option is to allow VPC CIDR block if other things in the VPC needs access)

### DIFF
--- a/packages/graphql-mesh-server/lib/maintenance.ts
+++ b/packages/graphql-mesh-server/lib/maintenance.ts
@@ -1,6 +1,6 @@
 import { Duration, RemovalPolicy } from "aws-cdk-lib";
 import * as apigateway from "aws-cdk-lib/aws-apigateway";
-import { IVpc, Peer, Port, SecurityGroup } from "aws-cdk-lib/aws-ec2";
+import { IVpc, SecurityGroup } from "aws-cdk-lib/aws-ec2";
 import * as lambda from "aws-cdk-lib/aws-lambda";
 import { NodejsFunction } from "aws-cdk-lib/aws-lambda-nodejs";
 import { Construct } from "constructs";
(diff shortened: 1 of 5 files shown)


packages/graphql-mesh-server/lib/maintenance.ts
```diff
@@ -72,10 +72,8 @@ export class Maintenance extends Construct {
       resources: [efsVolume.fileSystemArn, accessPoint.accessPointArn],
     });
 
-    efsVolumeSecGroup.addIngressRule(
-      Peer.anyIpv4(), // Can't get the IP address of each container as we don't know them at deploy time!
-      Port.tcp(2049),
-      "File access"
+    efsVolume.connections.allowDefaultPortFrom(
+      props.fargateService.connections
```
TheOrangePuff, line 76: I could not work this out when I wrote this!!! Thank you for fixing it :pray:

TheOrangePuff (approved):
Looks good!

packages/graphql-mesh-server/lib/maintenance.ts
gowrizrh: Actually, I think both the task and the function need to access the filesystem right? In that case it might be better to allow the VPC CIDR? (instead of Any IPv4)

gowrizrh: Or, create security groups for Lambdas and attach the SG to the ingress rule :thinking: what do you think?

gowrizrh: I've made it so that the vpc ipv4 addresses are allowed

gowrizrh:
@TheOrangePuff Ready for a re-review

* Added optional name prop for code pipeline service
* Use VPC IPv4 address
* Add authenticationTable prop to top level prop so it can be passed down (currently there is no way to customise this or disable it. Setting `''` can act as disable for now)

AdamJHall (approved):

PR #1358, opened 2024-05-08
chore: Add props to top level construct

gowrizrh:

* :facepalm: I forgot to use `serviceName` prop in the actual construct
* All the props added needed to be actually available at the top level construct so they can be passed down #1353


crispy101:
LGTM

crispy101 (approved):

PR #1364, opened 2024-05-09
fix: actually lock the version this time

TheOrangePuff:
Actually lock the version :grimacing:


finn-holland-aligent (approved):
:seal: of approval

PR #1367, opened 2024-05-13
DO-1652: Add name props, remove dependabot

gowrizrh:

* Removed dependabot
* Add optional name props

### DIFF
--- a/packages/prerender-fargate/lib/prerender-fargate.ts
+++ b/packages/prerender-fargate/lib/prerender-fargate.ts
@@ -119,7 +119,8 @@ export class PrerenderFargate extends Construct {
 
     const cluster = new ecs.Cluster(this, `${prerenderName}-cluster`, {
       vpc: vpc,
-      clusterName: props.clusterName !== undefined ? props.clusterName : undefined
+      clusterName:
+        props.clusterName !== undefined ? props.clusterName : undefined,
     });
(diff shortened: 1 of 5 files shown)


packages/prerender-fargate/lib/prerender-fargate.ts
```diff
@@ -119,6 +119,8 @@ export class PrerenderFargate extends Construct {
 
     const cluster = new ecs.Cluster(this, `${prerenderName}-cluster`, {
       vpc: vpc,
+      clusterName:
+        props.clusterName !== undefined ? props.clusterName : undefined,
```
finnholland, line 123: Might be better to have this as `props.clusterName ? props.clusterName : undefined` so that it equates null as false too along with the other ternaries.
feel free to ignore and merge

finnholland (approved):

TheOrangePuff (approved):

packages/prerender-fargate/lib/prerender-fargate.ts
gowrizrh: I think that makes sense

Although all (most?) props define `propName?` so only assigning the actual type or undefined is possible

Keeping this for consistency's sake as most other things check explicitly for undefined but we should reconsider this in a future major version

PR #1375, opened 2024-05-23
DO-1652: Add top level props for rest, queue names

gowrizrh:

* Added top level props so that the names can be passed to constructs


TheOrangePuff (approved):
LGTM

PR #1380, opened 2024-05-27
DO-1631: Use secrets manager for passing secrets

gowrizrh:

* Added a new prop for passing secrets via secrets manager. The ECS task execution will pass the secret value as a JSON object when launching the task.

Example:

image

### DIFF
--- a/packages/graphql-mesh-server/lib/fargate.ts
+++ b/packages/graphql-mesh-server/lib/fargate.ts
@@ -62,7 +62,9 @@ export interface MeshServiceProps {
    *
    * @deprecated - Use secrets instead
    */
-  ssmSecrets?: { [key: string]: ssm.IStringParameter | ssm.IStringListParameter };
+  ssmSecrets?: {
+    [key: string]: ssm.IStringParameter | ssm.IStringListParameter;
+  };
 
   /**
    * ECS Secrets to pass through to the container as secrets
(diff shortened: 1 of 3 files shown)


packages/graphql-mesh-server/lib/fargate.ts
```diff
@@ -61,6 +61,14 @@ export interface MeshServiceProps {
    * SSM values to pass through to the container as secrets
    */
   secrets?: { [key: string]: ssm.IStringParameter | ssm.IStringListParameter };
+
+  /**
+   * ECS Secrets to pass through to the container as secrets
+   *
+   * The key values can be referenced from either SSM or Secrets manager
+   */
+  secretsV2?: { [key: string]: ecs.Secret };
```
TheOrangePuff, line 70: Can we name this something else? `secretsName` or similar?

TheOrangePuff (changes requested):

packages/graphql-mesh-server/lib/fargate.ts
gowrizrh: This isn't expecting a single secret but a key value pair of secrets. I'd like to keep it so it's flexible (can take in multiple secrets if necessary) and also slightly clear that `secrets` and `secretsV2` are closely related and the newer approach should be preferred.

TheOrangePuff (approved):
LGTM

PR #1384, opened 2024-05-30
feat: add case insensitive flag

TheOrangePuff:

* Include case insensitive flag


finn-holland-aligent (approved):
seal3

PR #1385, opened 2024-05-31
Prerender structured logging

gowrizrh:

* Add structured logging for prerender. This allows us to store the logs, parse and have visibility on access.

image

### DIFF
--- a/packages/prerender-fargate/lib/prerender/Dockerfile
+++ b/packages/prerender-fargate/lib/prerender/Dockerfile
@@ -13,8 +13,8 @@ COPY ./package.json .
 
 RUN npm install --no-package-lock
 
-COPY ./server.js .
 COPY ./utils ./utils
+COPY ./server.js .
 
 EXPOSE 3000
 
--- a/packages/prerender-fargate/lib/prerender/Dockerfile
+++ b/packages/prerender-fargate/lib/prerender/Dockerfile
@@ -4,18 +4,17 @@ ENV CHROME_BIN=/usr/bin/chromium-browser
 ENV CHROME_PATH=/usr/lib/chromium/
 ENV MEMORY_CACHE=0
 
-# install chromium, tini and clear cache
-RUN apk add --update-cache chromium tini \
- && rm -rf /var/cache/apk/* /tmp/*
+RUN apk add --no-cache chromium tini
 
 USER node
 WORKDIR "/home/node"
 
 COPY ./package.json .
-COPY ./server.js .
 
-# install npm packages
 RUN npm install --no-package-lock
+
+COPY ./server.js .
+COPY ./utils ./utils
 
 EXPOSE 3000
 
--- a/packages/prerender-fargate/lib/prerender/server.js
+++ b/packages/prerender-fargate/lib/prerender/server.js
@@ -12,13 +12,48 @@
 'use strict';
 
 const prerender = require('prerender');
+const util = require('prerender/lib/util');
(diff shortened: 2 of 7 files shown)


packages/prerender-fargate/lib/prerender/Dockerfile
```diff
@@ -4,19 +4,18 @@ ENV CHROME_BIN=/usr/bin/chromium-browser
...
 WORKDIR "/home/node"
 
 COPY ./package.json .
-COPY ./server.js .
 
-# install npm packages
 RUN npm install --no-package-lock
 
+COPY ./server.js .
+COPY ./utils ./utils
```
TheOrangePuff, line 17: Minor but I think these should be swapped around, we're more likely to modify the server.js file

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -12,15 +12,50 @@
...
     forwardHeaders: true,
     chromeLocation: '/usr/bin/chromium-browser'
 });
 
+server.use({
+    beforeSend: (req, res, next) => {
+        const ms = new Date().getTime() - req.prerender.start.getTime();
+
+        if (req.prerender.url !== "health") {
+            logger.render({
```
TheOrangePuff, line 45: It's not clear what this is doing? I assume it only renders the logs if the req.prerender.url is not health? What about process logs? Startup, shutdown, etc?

TheOrangePuff (changes requested):
Couple questions but looks good!

packages/prerender-fargate/lib/prerender/server.js
gowrizrh: Because this is a server plugin on `beforeSend` this `render` log happens always before the server sends a response. And that means ELB healthchecks too.

`logger.render` is a custom log level that we would like to use to log crawler and bot requests.

Everything else is logged with `logger.info` as you can see in the PR's description screenshot

> image

packages/prerender-fargate/lib/prerender/Dockerfile
gowrizrh: Updated

gowrizrh:
@TheOrangePuff ready for another review :slightly_smiling_face:

packages/prerender-fargate/lib/prerender/server.js
gowrizrh: I've made a dedicated `/health` endpoint so that returns `200 OK` right away.

TheOrangePuff: Can you add a comment for that?

packages/prerender-fargate/lib/prerender/server.js
```diff
@@ -12,15 +12,59 @@
...
 const server = prerender({
     chromeFlags: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-web-security', '--remote-debugging-port=9222', '--hide-scrollbars', '--disable-dev-shm-usage'],
     forwardHeaders: true,
     chromeLocation: '/usr/bin/chromium-browser'
 });
 
+// Healthcheck endpoint
+server.use({
+    beforeSend: (req, res, next) => {
+        if (req.prerender.url === "health") {
```
TheOrangePuff, line 43: Oh wait nvm, this is good now :smile:

TheOrangePuff (approved):

TheOrangePuff (approved):

PR #1393, opened 2024-07-16
Hotfix: Allow multiple stacks

AdamJHall:
Update the mesh construct so that it can be deployed multiple times to the same account with the same defaults. Currently if you deploy it more than once it will fail and rollback due to creating a dashboard and dynamo DB table with the same name in both stacks. This PR removes the default dynamo DB table and updates the dashboard to include the stack name so it will not conflict.

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/graphql-mesh-server/lib/metrics.ts
+++ b/packages/graphql-mesh-server/lib/metrics.ts
@@ -30,6 +30,7 @@ export interface PerformanceMetricsProps {
   logGroup: LogGroup;
   snsTopic?: Topic;
   additionalAlarms?: Alarm[];
+  dashboardName?: string;
 }
 
 export class PerformanceMetrics extends Construct {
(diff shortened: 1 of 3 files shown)


packages/graphql-mesh-server/lib/metrics.ts
```diff
@@ -355,7 +356,8 @@ export class PerformanceMetrics extends Construct {
 
     // Create the dashboard
     new Dashboard(this, "dashboard", {
-      dashboardName: "Mesh-Dashboard",
+      dashboardName:
+        props.dashboardName ?? Stack.of(this).stackName + "-Mesh-Dashboard",
```
TheOrangePuff, line 360: I think defaulting to undefined is fine. CDK will just use an auto generated dashboard name

TheOrangePuff (changes requested):
Minor change but looks good!

packages/graphql-mesh-server/lib/metrics.ts
AdamJHall: Updated now :+1:

TheOrangePuff (approved):

PR #1396, opened 2024-09-09
DO-1704: add nginx to the mesh

TheOrangePuff:

* The mesh is currently missing access logs. The application layer logs are often unreliable too. By adding an nginx container that sits in front of the mesh we can ensure we are capturing access level logs.
* We want to ensure this works as an in place upgrade for existing stacks. By passing a custom task definition we can still use the same load balancer and therefore upgrade without changing DNS records.'

**Notes to reviewers**  

* My nginx config skills are not the best so please let me know if anything needs changing there :pray: 

### DIFF
--- a/packages/graphql-mesh-server/assets/nginx/nginx.conf
+++ b/packages/graphql-mesh-server/assets/nginx/nginx.conf
@@ -1,0 +1,53 @@
+events {
+  worker_connections 1024;
+}
+
+http {
+  log_format json_combined escape=json
+  '{'
+    '"time_local":"$time_local",'
+    '"remote_addr":"$remote_addr",'
+    '"remote_user":"$remote_user",'
+    '"request":"$request",'
+    '"status": "$status",'
+    '"body_bytes_sent":"$body_bytes_sent",'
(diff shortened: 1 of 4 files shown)


packages/graphql-mesh-server/assets/nginx/nginx.conf
```diff
@@ -0,0 +1,53 @@
+events {
+  worker_connections 1024;
+}
+
+http {
+  log_format json_combined escape=json
```
hubertott, line 6: JSON logging threw me, but this seems correct.

hubertott (approved):
❤️ NGINX

packages/graphql-mesh-server/assets/nginx/nginx.conf
TheOrangePuff: Yeah it looks a bit funky in the config but it seems to work. We want to use JSON so we have better CloudWatch query support: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_AnalyzeLogData-discoverable-fields.html

PR #1397, opened 2024-09-09
introduce logging to WAF, bumping up to 2.2.0

crispy101:

* introduce CloudWatch logging to WAF, keeping default as true. 

* 

* 

**Notes to PR author**

**Notes to reviewers**  

* log group name needs to be "aws-waf-logs-xxxxx" in order to be associated to a WebACL.

### DIFF
--- a/packages/waf/lib/waf.ts
+++ b/packages/waf/lib/waf.ts
@@ -79,7 +79,7 @@ export interface WebApplicationFirewallProps {
   postProcessCustomRules?: aws_wafv2.CfnWebACL.RuleProperty[];
 
   /**
-   * Enable CloudWatch logging. Default: false
+   * Enable CloudWatch logging. Default: true
    */
   enableLogging?: boolean;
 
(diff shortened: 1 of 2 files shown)


packages/waf/lib/waf.ts
```diff
@@ -76,6 +77,21 @@ export interface WebApplicationFirewallProps {
    * Priority numbers must be equal to or bigger than 30
    */
   postProcessCustomRules?: aws_wafv2.CfnWebACL.RuleProperty[];
+
+  /**
+   * Enable CloudWatch logging. Default: false
+   */
+  enableLogging?: boolean;
```
TheOrangePuff, line 84: Why not set this as true by default? I feel like logging as a default is a good thing

TheOrangePuff (changes requested):

packages/waf/lib/waf.ts
crispy101: Just to keep the status quo by default in terms of cost, I thought. But yeah that sounds better, as the stacks that are using older versions would need to bump up the construct version for future changes anyways. 
Updated.

TheOrangePuff (approved):
LGTM!

PR #1398, opened 2024-09-12
fix: specifically target mesh container for deploy

TheOrangePuff:

* When nginx is in use, this becomes the default container. However when the mesh deploys we want to target the mesh container. This removes the assumption that the mesh is the default container and instead searches for it.

**Notes to reviewers**  


crispy101 (approved):
LGTM

PR #1401, opened 2024-10-14
Behaviour Prefixes (for CDK v2 upgrade)

finnholland:

- Added an `id` tag to prerender function names to avoid duplicate named function issues
- Export remap path from construct for easier uniformity with types
- `behaviourPrefixes` to allow for behaviours on regional prefixes such as `au/` or `nz/`

**Notes to reviewers**  

- These changes allow for the CDK v2 upgrade to work for multiple brands, impact on other clients untested but it is all conditional so shouldn't affect them

### DIFF
--- a/packages/static-hosting/lib/static-hosting.ts
+++ b/packages/static-hosting/lib/static-hosting.ts
@@ -140,6 +140,13 @@ export interface StaticHostingProps {
   enableStaticFileRemap?: boolean;
 
   /**
+   * Any prefixes to remapping that should be included in the path such as au or nz
+   *
+   * @default true
+   */
+  behaviourPrefixes?: { prefix: string; edgeLambdas: EdgeLambda[] }[];
+
+  /**
    * Optional additional properties for static file remap behaviours
    *
    * @default none
@@ -292,7 +299,7 @@ export interface StaticHostingProps {
   comment?: string;
 }
 
-interface remapPath {
+export interface remapPath {
   from: string;
   to?: string;
   behaviour?: Partial<BehaviorOptions>;
@@ -559,6 +566,31 @@ export class StaticHosting extends Construct {
         };
       }
     }
+
+    if (enableStaticFileRemap) {
+      const staticFileRemapPrefixes = props.behaviourPrefixes?.map(
+        prefix => `${prefix.prefix}/`
+      ) || [""];
+      staticFileRemapPrefixes.forEach(prefix => {
+        this.staticFiles.forEach(path => {
+          additionalBehaviors[`${prefix}*.${path}`] = {
+            origin: s3Origin,
+            viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+          };
+        });
+      });
+    }
+
+    props.behaviourPrefixes?.forEach(prefix => {
+      additionalBehaviors[`${prefix.prefix}*`] = {
+        origin: s3Origin,
+        viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+        edgeLambdas: prefix.edgeLambdas,
+        originRequestPolicy: originRequestPolicy,
+        cachePolicy: originCachePolicy,
+        responseHeadersPolicy: responseHeadersPolicy,
+      };
+    });
 
     if (props.responseHeadersPolicies?.defaultBehaviorResponseHeaderPolicy) {
       defaultBehavior.responseHeadersPolicy =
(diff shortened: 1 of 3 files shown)


packages/static-hosting/lib/static-hosting.ts
```diff
@@ -139,6 +139,13 @@ export interface StaticHostingProps {
    */
   enableStaticFileRemap?: boolean;
 
+  /**
+   * Any prefixes to remapping that should be included in the path such as au or nz
+   *
+   * @default true
+   */
+  behaviourPrefixes?: { prefix: string; edgeLambdas: EdgeLambda[] }[];
```
TheOrangePuff, line 147: Not sure what the point of this is :thinking: We already have the `additionalBehaviors` prop and `staticFileRemapPrefixes` to handle prefixes. 

The name might just need updating to make it clear what it does

TheOrangePuff (changes requested):

packages/static-hosting/lib/static-hosting.ts
finnholland: updated this to what we discussed defaultBehaviourPrefixes

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -571,11 +574,11 @@ export class StaticHosting extends Construct {
       });
     }
 
-    props.behaviourPrefixes?.forEach(prefix => {
+    props.defaultBehaviourPrefixes?.forEach(prefix => {
       additionalBehaviors[`${prefix.prefix}*`] = {
         origin: s3Origin,
         viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
-        edgeLambdas: prefix.edgeLambdas,
+        edgeLambdas: prefix.behaviourOverride.edgeLambdas,
```
TheOrangePuff, line 581: Instead of just passing the edge lambda you can do this:

```typescript
      additionalBehaviors[`${prefix.prefix}*`] = {
        origin: s3Origin,
        viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
        originRequestPolicy: originRequestPolicy,
        cachePolicy: originCachePolicy,
        responseHeadersPolicy: responseHeadersPolicy,
        ...prefix.behaviourOverride,
      };
```
Because that will then allow you to override any of the properties if you need

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -140,11 +140,14 @@ export interface StaticHostingProps {
   enableStaticFileRemap?: boolean;
 
   /**
-   * Any prefixes to remapping that should be included in the path such as au or nz
+   * Overrides default behaviour paths with a prefix and takes in options to apply to each static file behaviour
```
TheOrangePuff, line 143: The options don't apply to the static file behaviour do they?

TheOrangePuff (changes requested):

packages/static-hosting/lib/static-hosting.ts
TheOrangePuff: @finn-holland-aligent this one hasn't been resolved yet

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -560,6 +560,31 @@ export class StaticHosting extends Construct {
       }
     }
 
+    if (enableStaticFileRemap) {
+      const staticFileRemapPrefixes = props.defaultBehaviourPrefixes?.map(
+        prefix => `${prefix.prefix}/`
+      ) || [""];
```
TheOrangePuff, line 566: Just to confirm you've tested this without passing in any default behaviour prefixes?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -560,6 +560,31 @@ export class StaticHosting extends Construct {
...
+        this.staticFiles.forEach(path => {
+          additionalBehaviors[`${prefix}*.${path}`] = {
+            origin: s3Origin,
+            viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
+          };
+        });
+      });
+    }
+
+    props.defaultBehaviourPrefixes?.forEach(prefix => {
```
TheOrangePuff, line 577: Pretty sure this won't work with a blank string :thinking: And if it did it wouldn't be the desired behaviour

TheOrangePuff (changes requested):

packages/static-hosting/lib/static-hosting.ts
TheOrangePuff: Never mind, I thought this was in the if statement :sweat_smile: If the user passes a blank string then that's on them haha

TheOrangePuff (approved):
LGTM

PR #1402, opened 2024-10-16
feat: remove legacy construct

TheOrangePuff:

* Remove the lambda at edge handlers construct as we don't use it anymore.

**Notes to reviewers**  


AdamJHall (approved):

AdamJHall (changes requested):
We need to remove from the README.md as well

AdamJHall (approved):

PR #1404, opened 2024-11-01
fix: add support for http2

TheOrangePuff:

Despite the AWS docs saying this value is quote: "Maximum HTTP version to support". This is a lie :-1: 

What it *actually* means is what the HTTP version should be with a fallback to version 1.1. If you want HTTP3 to fallback to HTTP2 you must explicitly define that. 

I can't think of a situation where we'd only want HTTP3 but not HTTP2 so I think this is a reasonable default.

**Notes to reviewers**  


crispy101 (approved):
LGTM

PR #1405, opened 2024-11-14
Chore/do 1587 migrate to yarn

TheOrangePuff:

This is a big change so apologies for the big PR :pray: Most of it is repeated config files for each package so hopefully it shouldn't take too long to review.

Converts our cdk-constructs to use yarn berry + nx for mono repo. I haven't changed any behaviour (at least not intentionally) so builds, publishing, etc should still work as before, just using a different bit of tech underneath.

I wanted to use PnP with yarn but this doesn't support nx project schemas :disappointed: (even though the nx docs say it does...).

**Notes to reviewers**  


.github/workflows/build.yml
```diff
@@ -4,16 +4,26 @@ on: [pull_request]
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
+
+      - name: Install
+        run: yarn install
+
+      - name: Build all packages
+        run: yarn build
```
AdamJHall, line 27: This should be `yarn nx run-many -t build` or the build script in the `package.json` should be, otherwise you aren't pulling any of the config defined in the `build` targets in each of the `project.json` files

.github/workflows/build.yml
```diff
@@ -4,16 +4,26 @@ on: [pull_request]
...
+
+      - run: yarn lint:check && yarn format:check
```
AdamJHall, line 29: Same with these it should be:
`yarn nx format:check` and `yarn nx run-many -t lint` OR  `yarn nx affected:lint --base=$DESTINATION_BRANCH --parallel --max-parallel=3` to only execute against changes on a PR

.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "semi": true,
   "singleQuote": false,
```
AdamJHall, line 3: We've got `singleQuote = true` on the internal NX repo, did we want to change one to match if so which one?

package.json
```diff
@@ -18,19 +18,37 @@
   },
   "homepage": "https://github.com/aligent/aws-cdk-constructs#readme",
   "scripts": {
-    "build": "tsc",
-    "lint": "eslint --ignore-path .eslintignore --ext .ts .",
-    "lint:check": "npm run lint",
-    "lint:fix": "npm run lint --fix",
+    "build": "nx exec -- tsc",
+    "lint": "eslint . --ignore-pattern '.pnp.*'",
```
AdamJHall, line 22: As above, we shouldn't call eslint, prettier etc ourselves but instead through the NX targets / NX format commands.

AdamJHall (changes requested):
Config looks good, just need to update the workflows and package.json scripts to go through the NX targets / executors so it picks up the config there and is cacheable.

AdamJHall (approved):
Looks good :)

PR #1414, opened 2024-12-05
Fix/include handlers in build

TheOrangePuff:

* NX uses the root .gitignore and .npmignore when building npm packages. So we have to write custom logic if we want to include ignore files in specific packages. This PR adds a script to handle that
* Removes all ignore files that weren't needed
* Configures NX to include additional required package assets 

**Notes to reviewers**  

When you've finished leaving feedback, please add a final comment to the PR tagging the author, letting them know that you have finished leaving feedback


AdamJHall (approved):
Looks good :)

PR #1421, opened 2025-01-19
GeoIP | Modular Region Domains and Rewrite Paths

finnholland:

* Created a new interface `RedirectFunctionOptions` to provide a better way to pass region codes and domains to redirect to.
* Changed the functional logic to allow for more than just path rewrites such as `.com/au`. You can now redirect directly to a site or to a path based rewrite.
* Added a README

**Notes to reviewers**  

### DIFF
--- a/packages/geoip-redirect/lib/handlers/redirect.ts
+++ b/packages/geoip-redirect/lib/handlers/redirect.ts
@@ -20,7 +20,7 @@ export const handler = async (
   options.defaultRegionCode = process.env.DEFAULT_REGION_CODE ?? "";
   const defaultRegion = options.defaultRegionCode.split("|")[0].toLowerCase();
   options.supportedRegions = {
-    ...process.env.SUPPORTED_REGIONS as Record<string, DomainOverwrite> ?? "{}",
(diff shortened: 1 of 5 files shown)


TheOrangePuff (changes requested):
Left a few comments. Main bit of feedback is it needs some more comments in the code to explain _why_ we're doing certain things

packages/geoip-redirect/lib/handlers/redirect.ts
```diff
@@ -5,34 +5,71 @@ import {
...
-
     return {
       status: "302",
       statusDescription: "Found",
       headers: {
         location: [
           {
             key: "Location",
-            value: redirectURL,
+            value: redirectURL.replace("/index.html", ""),
```
TheOrangePuff, line 72: Not sure why this is here?

TheOrangePuff (changes requested):

packages/geoip-redirect/lib/handlers/redirect.ts
finnholland: 🤷‍♂️ didn't seem to work without it but maybe just the minimum working example i was using 🤔

TheOrangePuff (approved):

PR #1422, opened 2025-01-22
Header change detection construct 

TheOrangePuff:

This is a big PR, apologies. Probably best to read the readme which describes the purpose of what I'm adding here. TL;DR: this is required for PCI DSS v4 11.6.1.

This creates a Lambda, DynamoDB table, and EventBridge schedule to check for header changes. The idea is that this construct can be added to any hosting stack to add the header change detection functionality.

**Notes to reviewers**  

@toddhainsworth I've added you specifically for the Lambda function code review but feel free to review other areas as well.


TheOrangePuff:
The build is failing for a strange reason :thinking: 

```
node_modules/@smithy/core/protocols.d.ts:6:3 - error TS2308: Module "@smithy/core/dist-types/submodules/protocols/index.d" has already exported a member named 'RequestBuilder'. Consider explicitly re-exporting to resolve the ambiguity.

6   export * from "@smithy/core/dist-types/submodules/protocols/index.d";
```

packages/header-change-detection/lib/header-change-detection.ts
```diff
@@ -0,0 +1,114 @@
...
+        bundling: {
+          command,
+          image: DockerImage.fromRegistry("busybox"),
+          local: new Esbuild({
+            entryPoints: [join(__dirname, "lambda/header-check.ts")],
+          })
+        }
+      }),
+      environment: {
+        "URLS": "https://production.pwa.aligent.com.au/,https://production.pwa.aligent.com.au/checkout",
```
AdamJHall, line 101: This should be coming through as a prop right?

packages/header-change-detection/jest.config.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */
+export default {
+  displayName: "basic-auth",
+  preset: "../../jest.preset.js",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.[tj]s$": ["ts-jest", { tsconfig: "<rootDir>/tsconfig.spec.json" }],
+  },
+  moduleFileExtensions: ["ts", "js", "html"],
+  coverageDirectory: "../../coverage/packages/basic-auth",
```
toddhainsworth, line 10: Should `basic-auth` here be `header-change-detection`?

packages/header-change-detection/lib/header-change-detection.ts
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    const schedule = new Rule(this, 'EventRule', {
+      schedule: props.schedule || Schedule.rate(Duration.hours(1)),
+      ...props.ruleProps
+    });
+
+    const lambda = new Function(this, 'HeaderCheck', {
+      architecture: Architecture.X86_64,
+      runtime: Runtime.NODEJS_20_X,
```
toddhainsworth, line 89: Didn't we need Node 22 for one of the `Map` iterators? Or did we move away from that?

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
+import axios from "axios";
+import { DynamoDBClient, BatchGetItemCommand, BatchGetItemCommandInput, KeysAndAttributes, UpdateItemCommandInput, UpdateItemCommand, AttributeValueUpdate } from "@aws-sdk/client-dynamodb";
```
toddhainsworth, line 2: Surprised eslint or prettier didn't split this up.

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+import { PublishCommand, PublishInput, SNSClient } from "@aws-sdk/client-sns";
+
+const URLS = process.env.URLS;
+const HEADERS = process.env.HEADERS;
+const TABLE = process.env.TABLE!;
```
toddhainsworth, line 7: We've been playing around in a couple of projects with using [env-var](https://github.com/evanshortiss/env-var) for erroring nicely when required environment variables are missed. Might be something to consider here, not essential.

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+
+const config = ""
+const DB_CLIENT = new DynamoDBClient(config);
+
+const securityHeaders = HEADERS?.split(",") || []
+
+type Headers = Map<string, string | undefined>
+type URLHeaders = Map<string, Headers>
```
toddhainsworth, line 15: I'm wondering if commentary describing what the string key is here (the URL, right?) might make it clearer?

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+    getStoredValues(urls),
+    fetchHeaders(urls)
+  ]);
+
+  // Find any differences between the headers
+  const headerDifferences = new Map() as Map<string, Difference[]>;
+  let differencesDetected = false;
+  const dbUpdates = urls.map(url => {
+    const currentHeaders = currentUrlHeaders.get(url)
+    const storedHeaders = storedUrlHeaders.get(url) || new Map()
```
toddhainsworth, line 31: Does TS infer the default `Map`'s internal type correctly here?
Would it be worth typing the default `Map` here so it's internals are known (if TS isn't able to infer from usage)
eg. `new Map<string, Headers>` or `new UrlHeaders()` if that's possible (not sure if TS will compile that back to the original type? :man_shrugging:

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    if (!currentHeaders) throw new Error(`Could not get current headers for ${url}`)
+
+    // Check all headers that we care about
+    headerDifferences.set(url, compareHeaders(securityHeaders, storedHeaders, currentHeaders));
+
+    const headersToUpdate: Headers = new Map();
```
toddhainsworth, line 38: Similar to the above, does `new Headers()` work here? More a curiosity than anything else :D

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+  if (differencesDetected)
+    await sendToSns(formatDifferences(headerDifferences));
+};
+
+/**
+ * Fetch security headers for the given urls
+ *
+ * @param urls list of urls to fetch headers from
+ */
+const fetchHeaders = async (urls: string[]) => {
```
toddhainsworth, line 58:
```suggestion
const fetchHeaders = async (urls: string[]): Promise<URLHeaders> => {
```

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+  const currentUrlHeaders: URLHeaders = new Map() as URLHeaders;
+
+  // Make an axios request for each url
+  await Promise.all(urls.map(url => axios.get(url).then(response => {
+    // Then get all the security headers from each response
+    const headers: Headers = new Map() as Headers;
```
toddhainsworth, line 64: Why does this one need an assertion but line 38 doesn't?

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+/**
+ * Update stored headers for the given url.
+ * If the header no longer has a value, delete it. Otherwise update it.
+ *
+ * @param url the url to update - this is the primary key
+ * @param headers record of headers to update
+ */
+const updateStoredValues = async (url: string, headers: Headers) => {
```
toddhainsworth, line 115:
```suggestion
const updateStoredValues = async (url: string, headers: Headers): Promise<UpdateItemCommandOutput | undefined> => {
```

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+  const batchGetInput: BatchGetItemCommandInput = { RequestItems: requestItems }
+  const batchGetCommand = new BatchGetItemCommand(batchGetInput);
+
+  // Fetch stored stored headers
+  const response = await DB_CLIENT.send(batchGetCommand);
+  const responses = response.Responses?.[TABLE]!;
+
+  console.log(`Got following data from dynamo table: ${JSON.stringify(responses)}`);
+
+  const storedUrlHeaders: URLHeaders = new Map() as URLHeaders;
```
toddhainsworth, line 163: Same here, having to `as ...` this feels weird. TS would be upset that we haven't provided generics for the Map, so we could just provide those and we won't have to persuade it.

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -0,0 +1,300 @@
...
+  return new Map([...storedUrlHeaders, ...nextUrlHeaders]);
+}
+
+/**
+ * Send an update command to DynamoDB
+ *
+ * @param url the url to update - this is the primary key
+ * @param attributes Record<string, AttributeValueUpdate>
+ */
+const dynamoUpdateRequest = async (url: string, attributes: Record<string, AttributeValueUpdate>) => {
```
toddhainsworth, line 191:
```suggestion
const dynamoUpdateRequest = async (url: string, attributes: Record<string, AttributeValueUpdate>): Promise<UpdateItemCommandOutput> => {
```

packages/header-change-detection/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@aligent/cdk-header-change-detection",
+  "version": "2.2.0",
+  "main": "index.js",
+  "license": "GPL-3.0-only",
+  "homepage": "https://github.com/aligent/aws-cdk-header-change-detection-stack#readme",
```
toddhainsworth, line 6: Should this (and the repository below) link back to the cdk-constructs repo?

packages/header-change-detection/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@types/jest": "^29.5.10",
+    "@types/node": "20.6.3",
+    "aws-cdk": "2.113.0",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.1.1",
+    "ts-node": "^10.9.1",
+    "typescript": "~5.3.2"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dynamodb": "^3.726.1",
```
toddhainsworth, line 25: Do we also need the SNS client?

toddhainsworth (changes requested):
Few little housekeeping things and TS questions - looking good!

packages/header-change-detection/lib/header-change-detection.ts
TheOrangePuff: Oops, yep!

packages/header-change-detection/lib/header-change-detection.ts
TheOrangePuff: I moved away from it because of this :sweat_smile: 
image

The version of cdk we have doesn't know that node 22 exists yet. We'd have to upgrade which would be a bigger effort. Something I'd like to get done in the future though

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: Yeah agree :thinking: but it doesn't seem to mind

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: I generally don't mind when passing stuff into Lambdas because unless something has gone terribly wrong it will always have a value passed in through the build process. It would be _better_ to have something to actually check it but I don't think it's that important here

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: You can't just call `new UrlHeaders()` because it's a type. I think if I rewrote it to extend the Map class it would work but that doesn't feel worth it imo.

You are right though, it doesn't know the type and sees it as `Map<any, any>`. I might use the `as` keyword here to make that slightly nicer.

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: Doesn't need one. Potentially I added it when the code looked different and needed one then. Just never removed it (but I'll leave it in as I'll add it to the others)

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: Oh I didn't realise you could do that with a map. I'll change the others too

toddhainsworth (dismissed):
LGTM! Nicely done :man_dancing:

AdamJHall (approved):

toddhainsworth (approved):

AdamJHall (approved):

toddhainsworth (approved):

PR #1434, opened 2025-01-31
CSP behaviour options

TheOrangePuff:

* Adds config to allow adding csp to specific paths. This will allow checkout to have it's own unique CSP. 
* Due to the length of our CSP headers, adding CSP requires the use of Lambda@Edge functions

**Notes to reviewers**  


packages/static-hosting/lib/csp.ts
```diff
@@ -0,0 +1,129 @@
...
+      define: {
+        "process.env.S3_BUCKET": JSON.stringify(options.bucket),
+        "process.env.CSP_OBJECT": JSON.stringify(
+          options.cspObject ?? "csp.txt"
+        ),
+        "process.env.REPORT_URI": JSON.stringify(options.reportUri ?? ""),
+        "process.env.FALLBACK_CSP": JSON.stringify(options.fallbackCsp ?? ""),
+      },
+      functionOptions: {
+        timeout: Duration.seconds(15),
```
AdamJHall, line 123: Should we drop the timeout here? 2-3 seconds should be enough?

packages/static-hosting/lib/handlers/csp-lambda/origin-response.ts
```diff
@@ -0,0 +1,73 @@
...
+const S3_BUCKET = process.env.S3_BUCKET;
+
+const REPORT_URI = process.env.REPORT_URI;
+
+const FALLBACK_CSP = process.env.FALLBACK_CSP;
+
+export const handler = async (
+  event: CloudFrontResponseEvent
+): Promise<CloudFrontResponse> => {
+  console.log("Lambda@Edge handler invoked");
```
AdamJHall, line 16: The logging here could get pretty loud, do we want to cut it back?

packages/static-hosting/lib/static-hosting.ts
```diff
@@ -547,6 +582,46 @@ export class StaticHosting extends Construct {
       }
     }
 
+    const cspPaths = props.cspPaths || [];
+    const cspRemapPaths = cspPaths.map(cspPath => {
+      const { path, indexPath, reportUri, fallbackCsp } = cspPath;
+
+      const requestFunction = new RequestFunction(
+        this,
+        "AlternativePathFunction",
```
AdamJHall, line 591: The names of constructs here will need the index of the array, otherwise if you have more than 1 element in `cspPaths` it will error as the construct already exists.

AdamJHall (changes requested):
Looks good, couple of small suggests and one change needed in the csp paths map.

AdamJHall (approved):
Looking good :sunglasses:

PR #1440, opened 2025-02-05
Use peer dependencies

TheOrangePuff:

* Change all our constructs to use peer dependencies and lock to a major version of cdk
* This should allow us to change the version of cdk in repos that use these constructs, rather than being forced to use the version the construct is written in

**Notes to reviewers**  


aaronmedina-dev (approved):
LGTM!

PR #1460, opened 2025-03-13
chore: fix template injection vulnerability

TheOrangePuff:

A GitHub Actions workflow step contains a template expression referencing potentially untrusted GitHub context fields. This may allow malicious input to be injected into shell commands, leading to a potential supply chain attack as tokens of the CI/CD pipeline could be exfiltrated.

The env block ensures GitHub Actions resolves the expression safely before passing it to the shell. This should avoid the above issue.

**Notes to reviewers**  

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
           fetch-depth: 0
 
       - name: Fetch target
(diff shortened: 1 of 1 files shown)


finnholland (approved):
thumbs-up-chuck-norris

PR #1464, opened 2025-03-23
Feature/DO-1834 Header Change Detection add HTTP status flags for custom behaviour for BigCommerce

aaronmedina-dev:
------

- Update Aligent's Header Change Detection to accommodate BigCommerce checkout behavior
- Added props to have an ability to add custom HTTP statuses.
- Joined the URL Array to a single string before saving.

Before the change
image

After the Change 
Test with added 406 status
image
Test without the 406
image

* 

**Notes to PR author**

**Notes to reviewers**  

### DIFF
--- a/packages/header-change-detection/lib/lambda/header-check.ts
+++ b/packages/header-change-detection/lib/lambda/header-check.ts
@@ -23,6 +23,13 @@ const DB_CLIENT = new DynamoDBClient(config);
 
 const securityHeaders = HEADERS?.split(",") || [];
 
+// Accept status 200 default
+const ACCEPTED_HTTP_STATUS = (process.env.ACCEPTED_HTTP_STATUS || "200")
+  .split(",")
+  .map(code => parseInt(code.trim(), 10));
+
+console.log("ACCEPTED_HTTP_STATUS:", ACCEPTED_HTTP_STATUS);
+
 type Headers = Map<string, string | undefined>;
 
 // A map of URLs and their headers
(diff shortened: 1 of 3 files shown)


packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -80,8 +87,14 @@ const fetchHeaders = async (urls: string[]): Promise<URLHeaders> => {
   // Make an axios request for each url
   await Promise.all(
     urls.map(url =>
-      axios.get(url).then(response => {
-        // Then get all the security headers from each response
+      axios.get(url, { validateStatus: () => true }).then(response => {
```
TheOrangePuff, line 90: Should we instead have `validateStatus` check the status codes?

I don't mind this approach but curious to know the thought process behind allowing all status codes to succeed and handling the status code check in the response

TheOrangePuff (changes requested):
Looks good! Just a question and the Aikido recommendation needs review :)

packages/header-change-detection/lib/lambda/header-check.ts
aaronmedina-dev: @TheOrangePuff 
I do not have specific reasons, I was just trying to minimise the code changes and add what's needed without overthinking it. I'm not yet familiar with axios as well so I might not have the best solution. But yep, using the axios error handling is another option and we just need to put a try-catch there.

Would you like me to update it to this implementation? :)

aaronmedina-dev:
> Looks good! Just a question and the Aikido recommendation needs review :)

I think for our use case, this is not applicable? Our URLs are passed as props to the construct at deploy time and not at runtime, so there's no user input that would allow SSRF exploitation (?). 
What do you think @TheOrangePuff ?

packages/header-change-detection/lib/lambda/header-check.ts
TheOrangePuff: Nah all good, I'm happy with this :)

packages/header-change-detection/lib/lambda/header-check.ts
```diff
@@ -80,8 +87,12 @@ const fetchHeaders = async (urls: string[]): Promise<URLHeaders> => {
...
```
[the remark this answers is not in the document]
aaronmedina-dev: Not applicable for our use case

TheOrangePuff (approved):

PR #1480, opened 2025-05-01
Removed set-cookie header, updated README

finnholland:

* removed `set-cookie` header from the monitoring so that it's less annoying (it's not a security header so not needed to monitor as per PCI standards)
* also updated the GeoIP readme as a small fix to due to changing the logic and forgot to change readme

**Notes to reviewers**  


crispy101 (approved):
LGTM

PR #1481, opened 2025-05-01
DO-1852: include cspObject property to specify CSP file

crispy101:

* include cspObject property to specify CSP file

**Notes to reviewers**  

### DIFF
--- a/packages/static-hosting/lib/static-hosting.ts
+++ b/packages/static-hosting/lib/static-hosting.ts
@@ -335,6 +335,11 @@ export interface CSPConfig {
    * be retrieved or parsed
    */
   fallbackCsp?: string;
+
+  /**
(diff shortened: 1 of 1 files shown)


aaronmedina-dev (approved):
LGTM.

Suggestion (not a deal breaker), we could consider a default fallback value for cspObject when it's not defined for readabilit?

crispy101:
> LGTM.
> 
> Suggestion (not a deal breaker), we could consider a default fallback value for cspObject when it's not defined for readabilit?

Sounds sensibel. 
Curretly, `cspObject` defaults to `csp.txt`. The `fallbackCsp` string is the fall-back CSP when the txt file is not found/readable. We can make `fallbackCsp` deafult to some string in this construct in case of everything missing.

PR #1489, opened 2025-05-30
TNT-1577: allow 4xx and 3xx to succeed

TheOrangePuff:

* Currently if Prerender returns a 4xx or 3xx axios will throw an error and the lambda will be marked as a failure. This places the page back onto the recache queue and this will repeat until the max retries have passed.
* Instead we should treat 4xx and 3xx as successes

**Notes to reviewers**  


AdamJHall (approved):

crispy101 (approved):
LGTM

PR #1518, opened 2025-07-28
TNT-1615: Fix CSP Lambda error handling when fallback is provided

TheOrangePuff:
## Summary
- Fixed error handling in CSP Lambda to properly use fallback CSP without throwing errors
- Lambda now returns successfully when fallback CSP is applied instead of returning error response

## Problem
The CSP Lambda was throwing an error even after successfully applying the fallback CSP value. This caused CloudFront to receive error responses despite the CSP headers being correctly set with the fallback value.

## Solution
Removed the `throw error` statement after setting the fallback CSP headers. The Lambda now:
1. Logs the error for debugging purposes
2. Applies the fallback CSP to response headers
3. Returns successfully without throwing

## Test plan
- [x] Deploy Lambda with fallback CSP configured
- [x] Remove/rename CSP file in S3 to trigger fallback
- [x] Verify CloudFront returns 200 with fallback CSP headers
- [x] Check Lambda logs show fallback was used

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/short-ways-tie.md
+++ b/.changeset/short-ways-tie.md
@@ -1,0 +1,5 @@
+---
+"@aligent/cdk-static-hosting": patch
+---
+
+Fix CSP Lambda error handling when fallback is provided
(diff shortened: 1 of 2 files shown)


crispy101 (approved):
LGTM

PR #1667, opened 2026-05-03
feat: add SecureRestApi construct for API Gateway REST APIs with API key authentication and usage plan throttling

toddhainsworth:

* Adds new `@aligent/cdk-secure-rest-api` package providing a `SecureRestApi` CDK construct
* Creates an API Gateway REST API with API key authentication (`apiKeyRequired: true`) and usage plan throttling on all routes
* Routes accept any CDK `Integration` (Lambda, HTTP, Mock, Step Functions, etc.) rather than hardcoding a specific integration type, giving callers full control over how endpoints are wired
* Configurable CORS preflight options, throttle rate/burst limits, and name overrides for the API key and usage plan
* Adds the construct to the root README

**Notes to reviewers**

ℹ️ When you've finished leaving feedback, please add a final comment to the PR tagging the author, letting them 
know that you have finished leaving feedback

**Toggl**

`AP21INT-28: Code Review API`

### DIFF
--- a/packages/constructs/secure-rest-api/lib/secure-rest-api.ts
+++ b/packages/constructs/secure-rest-api/lib/secure-rest-api.ts
@@ -1,0 +1,121 @@
+import {
+  ApiKey,
+  Cors,
+  Integration,
+  IRestApi,
+  RestApi,
+  UsagePlan,
+} from "aws-cdk-lib/aws-apigateway";
+import { HttpMethod } from "aws-cdk-lib/aws-apigatewayv2";
+import { Construct } from "constructs";
+
+export interface SecureRestApiRoute {
+  path: string;
+  methods: HttpMethod[];
+  integration: Integration;
+}
+
+export interface SecureRestApiProps {
+  /**
+   * The name of the API.
+   */
+  apiName: string;
+
+  /**
+   * Description for the API.
+   */
+  description?: string;
+
+  /**
+   * CORS configuration.
+   * @default Allows all origins, GET/OPTIONS methods, Content-Type and X-Api-Key headers
+   */
+  corsOptions?: {
+    allowOrigins?: string[];
+    allowMethods?: string[];
+    allowHeaders?: string[];
+  };
+
+  /**
+   * Routes to register on the API.
+   */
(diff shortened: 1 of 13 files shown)


packages/constructs/secure-rest-api/lib/secure-rest-api.ts
```diff
@@ -0,0 +1,121 @@
...
+      usagePlanName,
+    } = props;
+
+    this.api = new RestApi(this, "Api", {
+      restApiName: apiName,
+      description: description ?? `REST API for ${apiName} service`,
+      defaultCorsPreflightOptions: {
+        allowOrigins: corsOptions?.allowOrigins ?? Cors.ALL_ORIGINS,
+        allowMethods: corsOptions?.allowMethods ?? ["GET", "OPTIONS"],
+        allowHeaders: corsOptions?.allowHeaders ?? [
```
daniel-vdp: Is there ever a situation where you wouldn't want these headers? Wondering if it's worth changing these to `additionalHeaders` and appending to the list? I feel like majority of the time you'll just want to append. Maybe same with methods as well

packages/constructs/secure-rest-api/lib/secure-rest-api.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  constructor(scope: Construct, id: string, props: SecureRestApiProps) {
+    super(scope, id);
+
+    const {
+      apiName,
+      description,
+      corsOptions,
+      routes,
+      throttle = { rateLimit: 100, burstLimit: 200 },
```
daniel-vdp, line 82: I had no idea you could set defaults like this, very cool!

TheOrangePuff (dismissed):

packages/constructs/secure-rest-api/lib/secure-rest-api.ts
toddhainsworth: Yeah that's a good point, I'll have these appended

kai-nguyen-aligent (approved):
LGTM!

TheOrangePuff (approved):

PR #1680, opened 2026-05-10
fix: skip CDK-managed singletons in MicroserviceChecks

toddhainsworth:
## Summary
- `MicroserviceChecks` now short-circuits on construct paths owned by the CDK framework so memory size, timeout, tracing, and log retention nags no longer fire on resources consumers cannot configure.
- Covered path fragments: `Custom::CDKBucketDeployment` (aws-s3-deployment), `LogRetention`, `Custom::S3AutoDeleteObjects`, and the `cr.Provider` framework handler (`AWS679f53fac002430cb0da5b7982bd2287`).
- Consumers can retire bespoke `BucketDeploymentNagSuppressionAspect`-style workarounds.

## Why
Surfacing nags on framework-owned singletons is noise — there is no user-tunable knob behind them. Filtering at the pack source avoids polluting the synth output with suppression annotations.

## Changeset
`minor` bump on `@aligent/cdk-aspects` (rule output changes; defensible as `patch` since the prior nags were false positives — happy to downgrade if preferred).

## Test plan
- [x] `yarn nx test cdk-aspects` — 11/11 pass, including a new `CDK-managed singletons` block that synthesises a `BucketDeployment` and asserts `hasNoError` for memory/timeout/tracing on its singleton lambda.
- [x] `yarn nx lint cdk-aspects` — clean (pre-existing warnings in `dynamodb.ts` only).
- [x] Sanity check in a downstream consumer that previously needed a manual suppression aspect.

### DIFF
--- a/.changeset/microservice-checks-skip-cdk-managed.md
+++ b/.changeset/microservice-checks-skip-cdk-managed.md
@@ -1,0 +1,5 @@
+---
+"@aligent/cdk-aspects": minor
+---
+
+`MicroserviceChecks` now skips CDK-managed singleton resources (BucketDeployment custom handler, `LogRetention`, `Custom::S3AutoDeleteObjects`, and `cr.Provider` framework lambdas). These nodes have framework-owned runtime config that consumers cannot tune, so surfacing memory size, timeout, tracing, and log retention nags on them is noise. Consumers no longer need bespoke suppression aspects for these resources.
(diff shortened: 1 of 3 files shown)


kai-nguyen-aligent (approved):
Thanks, Todd. My initial thought on this was adding a "exclude" configuration so consumer can use base on their need but this looks much nicer and reduce friction on consumer side.

daniel-vdp (approved):

PR #1702, opened 2026-06-04
fix: skip CDK-managed singletons in ResourcePrefixAspect

toddhainsworth:
## Summary

`ResourcePrefixAspect` assigned deterministic, logical-id-derived physical names to CDK-managed singleton/framework resources — notably the `BucketDeployment` custom-resource handler and its nested IAM service role and log group. Because these singletons share a **fixed logical id across every stack**, the prefixed name collapses to a single value per prefix scope, removing the per-stack and per-creation uniqueness CloudFormation normally provides. This caused two distinct deploy failures.

### Failure mode 1 (severe): cross-stack IAM role collision

Two `BucketDeployment`-using stacks in the same stage/brand received the **same** account-global `RoleName`, so the second stack failed:

```
CREATE_FAILED | AWS::IAM::Role | ...CustomCDKBucketDeployment8693-... already exists
```

Version-independent (the aspect has set explicit `RoleName`/`FunctionName` on these singletons since ≥ 0.5.7).

### Failure mode 2: orphaned `/aws/lambda` LogGroup collision on retry

The handler runs during deploy; the Lambda service auto-creates `/aws/lambda/<fn>`. A failed/rolled-back deploy orphans that group (CFN never owned it, so `cdk destroy` doesn't clear it), and the deterministic function name regenerates the identical name on retry:

```
Resource of type 'AWS::Logs::LogGroup' with identifier
'.../aws/lambda/...-CustomCDKBucketDeployment8693-<hash>' already exists.
```

Surfaced after `0.6.1` (which correctly stopped mangling `/aws/` log group names — that fix exposed the latent collision).

## Fix

- Extract the singleton path-fragment list + `isCdkManagedSingleton` helper into a new internal `lib/cdk-managed-singletons.ts` — single source of truth.
- `MicroserviceChecks` refactored to consume it (behaviour-preserving).
- `ResourcePrefixAspect.visit()` skips these resources, leaving their physical names CloudFormation-generated. The path-`includes` skip covers the handler **and its nested children** (IAM service role, log group), resolving both failure modes at once. Keeps the `0.6.1` `/aws/` log-group fix intact.

Covered singletons: `Custom::CDKBucketDeployment`, `LogRetention`, `Custom::S3AutoDeleteObjects`, `cr.Provider` (`AWS679f53fac002430cb0da5b7982bd2287`).

## Testing

- TDD (red → green). Behaviour tests assert the `BucketDeployment` handler lambda, its IAM **service role**, and the `S3AutoDeleteObjects` handler all keep CloudFormation-generated names.
- Verified against the unfixed aspect that the role/function were previously pinned to `myapp-CustomCDKBucketDeployment8693-...`.
- `build` + `test` (58 passed) + `lint` all green.

## Changeset

`@aligent/cdk-aspects`: **patch**

### DIFF
--- a/.changeset/resource-prefix-skip-cdk-managed-singletons.md
+++ b/.changeset/resource-prefix-skip-cdk-managed-singletons.md
@@ -2,4 +2,11 @@
 "@aligent/cdk-aspects": patch
 ---
 
-Fix `ResourcePrefixAspect` to skip CDK-managed singleton/framework resources (`BucketDeployment` handler, `LogRetention`, `S3AutoDeleteObjects`, and `cr.Provider` framework lambdas) instead of pinning deterministic physical names onto them. Pinning a deterministic name removed CloudFormation's random-suffix orphan-avoidance, so a failed deploy could orphan the service-created `/aws/lambda/<fn>` log group and wedge every subsequent deploy on `AWS::Logs::LogGroup ... already exists`. The aspect now leaves these resources CloudFormation-named, matching the singleton-skip behaviour `MicroserviceChecks` already applies.
+Fix `ResourcePrefixAspect` to skip CDK-managed singleton/framework resources (`BucketDeployment` handler, `LogRetention`, `S3AutoDeleteObjects`, and `cr.Provider` framework lambdas) and their nested children (IAM service roles, log groups) instead of pinning deterministic physical names onto them.
+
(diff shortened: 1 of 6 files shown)


kai-nguyen-aligent (approved):
Wow, you're faster than me, Todd. I had this in my list but you beat me ❤️

AdamJHall (approved):

PR #1703, opened 2026-06-05
fix: isolate maintenance handler test filesystem state

toddhainsworth:
## Problem

`graphql-mesh-server`'s maintenance handler tests are flaky. They surfaced in the [release workflow run #26996333785](https://github.com/aligent/cdk-constructs/actions/runs/26996333785/job/79666741299), where `whitelist.test.ts › should handle GET method` returned `{ whitelist: [] }` instead of the seeded allowlist. A re-run of that same workflow now passes — confirming it was non-deterministic, not a real regression.

## Root cause

`jest.mock-env.ts` sets `MAINTENANCE_FILE_PATH = cwd()` for every suite, and `lib/file.ts` reads/writes `${MAINTENANCE_FILE_PATH}/maintenance.{enabled,disabled}`. So `whitelist.test.ts` and `maintenance.test.ts` operate on **one shared file**. Jest runs test files in parallel workers, so their `beforeEach` blocks (each `rmSync` + rewrite the file) race across processes:

1. `whitelist` writes the file with the seeded allowlist
2. `maintenance` (other worker) removes it and recreates it as `{ whitelist: [], sites: ... }`
3. `whitelist`'s GET reads the file → empty → assertion fails

A torn read (one worker reading mid-write by the other) is a second signature of the same race.

## Fix

- `lib/file.ts` now resolves `MAINTENANCE_FILE_PATH` **lazily** (`getPaths()`) instead of freezing it into module-level `const`s at import time, so each suite can point it at its own directory. It also throws a clear error if the path is unset rather than silently writing to `undefined/maintenance.*`.
- `whitelist.test.ts` and `maintenance.test.ts` each create a unique `mkdtempSync` directory and set `MAINTENANCE_FILE_PATH` to it; `afterAll` removes it. No shared state between suites.
- New `lib/file.test.ts` locks in lazy resolution + directory isolation as a regression guard.

## Verification

- Package suite run 8× consecutively: 8/8 green (was ~1-in-5 failing before).
- `nx lint graphql-mesh-server`: clean.

Includes a patch changeset for `@aligent/cdk-graphql-mesh-server`.

### DIFF
--- a/.changeset/isolate-maintenance-test-filesystem-state.md
+++ b/.changeset/isolate-maintenance-test-filesystem-state.md
@@ -1,0 +1,5 @@
+---
+"@aligent/cdk-graphql-mesh-server": patch
+---
+
+Resolve MAINTENANCE_FILE_PATH lazily so the maintenance handler honours the runtime-configured path; fixes flaky maintenance handler tests via per-suite filesystem isolation.
--- a/packages/constructs/graphql-mesh-server/assets/handlers/maintenance/lib/file.test.ts
+++ b/packages/constructs/graphql-mesh-server/assets/handlers/maintenance/lib/file.test.ts
@@ -1,0 +1,27 @@
+import { mkdtempSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { getFilePath, getMaintenanceFile, setWhitelist } from "./file";
+
+describe("maintenance file storage", () => {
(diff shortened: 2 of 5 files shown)


toddhainsworth:
hey @crispy101 + @AdamJHall I just noticed this is still open, do either of you have some time this week to give it a review?

AdamJHall (approved):

PR #1737, opened 2026-07-22
feat: respect explicit DynamoDB throughput in DynamoDbDefaultsAspect

toddhainsworth:
## Summary

`DynamoDbDefaultsAspect` treated its profile throughput as an **override** — it stamped `onDemandThroughput` / `provisionedThroughput` onto every table regardless of what the `Table` construct already declared. Under the `MEDIUM` profile this capped every on-demand table at 100/100 RRU/WRU with no per-table escape, throttling high-write tables (see the originating incident: a product-sync burst hitting `TableWriteMaxOnDemandThroughputExceeded`).

This change treats profile throughput as a **default, not an override**, mirroring how `billingMode` is already handled (`?? default`). A value explicitly set on the `Table` construct is respected; the profile only fills in tables that say nothing — so the staging cost cap stays intact for the default case.

## Changes

- `lib/defaults/dynamodb.ts` — only stamp throughput when the `CfnTable` hasn't already set it (`onDemandThroughput === undefined` / `provisionedThroughput === undefined`), applied symmetrically to both branches. Added a comment explaining the two-predicate condition.
- `lib/defaults/dynamodb.test.ts` — flipped the two "overrides existing" tests to "respects explicitly set" behaviour, exercised through the real L2 `Table` props path.
- `README.md` — added the (previously missing) DynamoDB Defaults section documenting the profiles and the respect-explicit behaviour.
- Changeset: **minor**.

## Consumer note

To lift the cap on a table, set the throughput **and** `billingMode: PAY_PER_REQUEST` on the `Table` — CDK's L2 discards the request-unit props when billing mode resolves to `PROVISIONED`, *before any aspect runs*. Documented in the README example.

## Downstream impact

Lets `gra-ap21` drop its `UncapOnDemandThroughputAspect` workaround and set throughput directly on the affected tables, then bump `@aligent/cdk-aspects`.

## Testing

- `yarn nx test cdk-aspects` — 67 passed
- `yarn nx lint cdk-aspects` — clean (2 pre-existing non-null-assertion warnings, untouched)

### DIFF
--- a/.changeset/respect-explicit-dynamodb-throughput.md
+++ b/.changeset/respect-explicit-dynamodb-throughput.md
@@ -1,0 +1,5 @@
+---
+"@aligent/cdk-aspects": minor
+---
+
+`DynamoDbDefaultsAspect` now treats profile throughput as a default rather than an override. Read/write throughput explicitly set on a `Table` construct is respected instead of being clobbered by the profile, so a high-throughput table can lift the MEDIUM on-demand cap (or set its own provisioned capacity) via normal construct props.
--- a/packages/cdk-aspects/README.md
+++ b/packages/cdk-aspects/README.md
@@ -81,6 +81,38 @@ const app = new App();
 Aspects.of(app).add(new StepFunctionsDefaultsAspect());
 ```
 
+### DynamoDB Defaults
+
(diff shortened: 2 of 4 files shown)


kai-nguyen-aligent (approved):
LGTM!

tmthrgd-aligent (approved):

PR #1744, opened 2026-07-31
Revert "[Aikido] Fix security issue in yargs via major version upgrade from 17.7.2 to 18.1.0"

AdamJHall:
Reverts aligent/cdk-constructs#1743


tmthrgd-aligent (approved):
Change looks fine. Why the need for the revert?

AdamJHall:
> Change looks fine. Why the need for the revert?

I thought it was passing because the last run on the workflows was green, but it was actually failing the builds. 

Reverting for now so it still builds and can release.

